PR #29, opened 2021-07-02
Update keywords

RustanLeino:
Four of the keywords included in this file are contextual. In particular,
```
least predicate
greatest predicate
least lemma
greatest lemma
```
are _keyword phrases_ in Dafny, but `least` and `greatest` are not keywords by themselves. Also, `to` and `downto` are keywords in the context of a `for` statement, but are not reserved elsewhere. Does this grammar have the power to make these distinctions? (If not, `least`, `greatest`, `to`, and `downto` are probably rare enough that it won’t bother too many users.)

Some operators are syntactically a prefix of others. For example, `<=` and `<==` and `<==>`. Do these need to be ordered in a particular way in this file?

### DIFF
--- a/syntaxes/Dafny.tmLanguage
+++ b/syntaxes/Dafny.tmLanguage
@@ -37,7 +37,7 @@
 		</dict>
 		<dict>
 			<key>match</key>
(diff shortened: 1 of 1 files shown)


camrein:
Thanks for the PR. Unfortunately, I cannot question your answers as I have never worked with this file format so far. But I have quickly checked the documentation of VSCode regarding [syntax highlighting](https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide) and the one of TextMate about [language grammar](https://macromates.com/manual/en/language_grammars). I think we could add support for these keyword phrases, but we may have to rework the file (I think polishing wouldn't be too bad considering that there weren't many changes to the file in the past 8 years).

About `to` and `downto`: It might be possible in combination with the [rule keys](https://macromates.com/manual/en/language_grammars#rule_keys) `begin` and `end`. Alternatively, we could enhance the language server with [semantic highlighting](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide) to support contextual keywords.

Unfortunately, I haven't found anything about cases where keywords (or in this case the operators) are prefixes of others.

PR #55, opened 2021-10-05
Add build instructions to the README

### LINKED ISSUE #54: Instructions for contributors/building locally
I've mostly been able to figure this out by following https://code.visualstudio.com/api/get-started/your-first-extension and reverse-engineerings the https://github.com/dafny-lang/ide-vscode/blob/master/.github/workflows/ci.yaml file, but it would be great to provide a few lines of simple instructions for anyone wanting to contribute or try unreleased changes locally.

camrein:
Resolves #54 

@robin-aws Do you have any further instructions that could be beneficial here?

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -1,0 +1,11 @@
+.github/**
+.vscode/**
+.vscode-test/**
+src/**
+.gitignore
+.yarnrc
+vsc-extension-quickstart.md
+**/tsconfig.json
+**/.eslintrc.json
+**/*.map
+**/*.ts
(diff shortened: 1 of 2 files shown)


robin-aws:
Thanks @camrein, this is definitely helpful!

Do you find that there is any performance difference between the debugging method vs. building and installing the VSIX file? In other words, if someone just wants to use the latest version to write Dafny rather than contribute to the plugin itself, is the latter approach worth the extra effort?

camrein:
If you start the extension without attaching the debugger (CTRL+F5), I don't think that you'll have any noticeable performance impact, if at all. Especially because most of the action happens within the language server.

Issue #61, opened 2021-10-21
Test for range of compatible language server versions in CI


robin-aws:
I've been doing something of a post-mortem on the recent Dafny issue where a Dafny user had a really rough time using the newest plugin with existing Dafny code: https://github.com/dafny-lang/dafny/issues/1426.

One of the big takeaways from it so far is that the plugin should make it easier to downgrade to an older version of Dafny, and in fact ensure we know what the minimum compatible version of the server is. 

In my experience, IDEs usually prompt you to configure the programming language SDK/runtime/etc. whenever you create or open a project for the first time, and it usually asks you to download/install them yourself and then configure where they are in the IDE. This makes it a much more explicit decision what version to use, and tends to persist that decision on a per-project rather than IDE-global way. Tweaking the UX in this workflow might also help users realize they should try a different Dafny version when things go awry.

@camrein: I hugely appreciate your jumping on the issues I've been cutting, even the ones that are just half-baked ideas. :) But this one is a lot more urgent in my opinion, in order to do what we can to avoid more users having suboptimal experiences. I'm hesitant to move to the full Dafny 3.3.0 release until we have some good mutual mechanisms in place, to make sure it's easy for users to keep getting plugin releases with all the great improvements you keep making, but to also make sure they can safely stay on the Dafny version currently working for them.

camrein:
> I've been doing something of a post-mortem on the recent Dafny issue where a Dafny user had a really rough time using the newest plugin with existing Dafny code: [dafny-lang/dafny#1426](https://github.com/dafny-lang/dafny/issues/1426).

This issue was indeed unpleasant. This may drive again the necessity to be able to release language server updates more independently from Dafny releases, or increase the release cycle. Most of the issues we received were not related to the extension but the language server. Alternatively, we might consider something like a nightly build to not have users compile Dafny manually in such urgent cases (or have us provide the binaries like in the issue above).

> One of the big takeaways from it so far is that the plugin should make it easier to downgrade to an older version of Dafny, and in fact ensure we know what the minimum compatible version of the server is.

Just to make sure we're talking about the same. This extension is only compatible to Dafny 3 onwards since it requires the LSP language server. The previous extension for Dafny 2 relied on DafnyServer which works differently. So it's not possible to downgrade prior Dafny 3 with this extension.
[...]

PR #90, opened 2021-12-01
fix: Open SaveAs for untitled documents

### LINKED ISSUE #89: Compile and Run on an unsaved editor closes it and does not run
ide-vscode-new-issue

Steps to reproduce this problem

- File > new file
- Select a language (Dafny)
- Paste some code like this:
```
method Main() {
  print "ok";
}
```
- Right-click, compile and run (or press F5)
- Give any name on any path, press save.

**Problem#1**: It displays this in the embedded terminal
```
PS C:\Users\.....> & "C:\Program Files\dotnet\dotnet.EXE" "c:\Users\.....\.vscode\extensions\correctnesslab.dafny-vscode-2.0.2\out\resources\dafny\Dafny.dll" "Untitled-1" /verifyAllModules /spillTargetCode:1 /compile:3 /out:bin\Untitled-1
*** Error: 'Untitled-1': Filename extension '' is not supported. Input files must be Dafny programs (.dfy) or C# files (.cs) or managed DLLS (.dll)
```
which is using `Untitled-1` instead of the newly entered file name.

**Problem#2**: It closes the tab containing the code, and offers no way to reopen it.

The file is correctly saved though.

camrein:
resolves #89

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 2.1.0
 
 - Now marking ghost statements by default
+- Fixed issue when compiling untitled documents
 
 ## 2.0.2
 
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -8,5 +8,6 @@ export namespace DafnyCommands {
 
 export namespace VSCodeCommands {
   export const Open = 'vscode.open';
+  export const SaveAs = 'workbench.action.files.saveAs';
   export const ConfigureLanguageSettings = 'workbench.action.configureLanguageBasedSettings';
 }
--- a/src/ui/compileCommands.ts
+++ b/src/ui/compileCommands.ts
@@ -2,7 +2,7 @@ import * as os from 'os';
 import * as path from 'path';
 
 import { window, commands, ExtensionContext } from 'vscode';
-import { DafnyCommands } from '../commands';
+import { DafnyCommands, VSCodeCommands } from '../commands';
 
 import Configuration from '../configuration';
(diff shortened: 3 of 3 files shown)


keyboardDrummer:
Do you think it would be useful to add a test? I sort of understand why you didn't, so I'll leave it up to you.

keyboardDrummer (approved):

camrein:
> Do you think it would be useful to add a test?

Since I'm a strong supporter of automized tests I'd always go for tests.

As of now, I'm having trouble setting up reliable UI tests and [VSCode's documentation](https://code.visualstudio.com/api/working-with-extensions/testing-extension) only illustrates a very basic example.
I've already tried to create a foundation for testing because of #61, but somehow the extension fails somewhere during initialization without VSCode providing an error (try...catch didn't help either). I've got the feeling that it's because of the OS/process interactions, but I haven't had the capacity to dive into that.

keyboardDrummer:
> > Do you think it would be useful to add a test?
> 
> Since I'm a strong supporter of automized tests I'd always go for tests.
> 
> As of now, I'm having trouble setting up reliable UI tests and [VSCode's documentation](https://code.visualstudio.com/api/working-with-extensions/testing-extension) only illustrates a very basic example. I've already tried to create a foundation for testing because of #61, but somehow the extension fails somewhere during initialization without VSCode providing an error (try...catch didn't help either). I've got the feeling that it's because of the OS/process interactions, but I haven't had the capacity to dive into that.

Sorry to hear that.

PR #91, opened 2021-12-01
Misleading message for compilation fixed

MikaelMayer:
A regular Dafny user pointed out that the compilation message is misleading for several reasons.

Current message:

    '$(thumbsup) Compilation succeeded';

Suggested message:

    'Resolution succeeded. Awaiting verification...';

Here is why:

- "Compilation" is normally the step that happens after verification. Hence, the current messages would indicate that Dafny files were translated before verification even started, which is not the case. Hence, I use the wording "Resolution" which is consistent with the other message `$(thumbsdown) Resolution Failed';`
- I removed $(thumbsup). Until recently, thumbs up was used only when verification succeeded, which makes sense given its "all right" meaning. Providing an intermediate thumbs up at resolution time feels totally unnecessary and even misleading to someone who might think for a second that verification succeeded. We could have an intermediate icon instead, but we should not assign the same icon to two different steps of the process.

### DIFF
--- a/src/ui/messages.ts
+++ b/src/ui/messages.ts
@@ -2,7 +2,7 @@ export namespace Messages {
   export namespace CompilationStatus {
     export const ParsingFailed = '$(thumbsdown) Parsing Failed';
     export const ResolutionFailed = '$(thumbsdown) Resolution Failed';
-    export const CompilationSucceeded = 'Resolution succeeded. Awaiting verification...';
+    export const CompilationSucceeded = '$(book) Resolved (not verified)';
     export const Verifying = '$(sync~spin) Verifying...';
     export const VerificationSucceeded = '$(thumbsup) Verification Succeeded';
     export const VerificationFailed = '$(thumbsdown) Verification Failed';
--- a/src/ui/messages.ts
+++ b/src/ui/messages.ts
@@ -2,7 +2,7 @@ export namespace Messages {
   export namespace CompilationStatus {
     export const ParsingFailed = '$(thumbsdown) Parsing Failed';
     export const ResolutionFailed = '$(thumbsdown) Resolution Failed';
-    export const CompilationSucceeded = '$(thumbsup) Compilation Succeeded';
+    export const CompilationSucceeded = 'Resolution succeeded. Awaiting verification...';
     export const Verifying = '$(sync~spin) Verifying...';
     export const VerificationSucceeded = '$(thumbsup) Verification Succeeded';
     export const VerificationFailed = '$(thumbsdown) Verification Failed';


seebees:
LGTM

camrein:
Thanks for this proposal. Your motivation makes sense to me and I'm happy to merge it.
However, I'm afraid of the term becoming too long. If people have lots of extensions installed, the space is becoming tight in the status bar. Do you have an idea for something shorter (I already had the feeling that the messages we have now are a bit long)? For example `Resolution succeeded (not verified)`. Nevertheless, we could await user feedback before thinking too long about this.

But for continuity of the status messages, I propose to add an icon since all status messages do have one. A list of the available icons can be found [here](https://code.visualstudio.com/api/references/icons-in-labels).
I could think of one of those:

- gear
- lightbulb (to indicate that the verification is pending)

MikaelMayer:
I understand your concern.
I also understand that verification might not happen right after resolution - resolution happens all the time almost on every keystroke, whereas users might enable verification only on save. Hence it is probably not appropriate to say that "Awaiting verification..." because there might be nothing to wait on until the user saves.
Down the road, I believe we could have another message in the case when document has to be verified after resolution. In that case, we should replace the message "Compilation succeeded" by just the message "Verification started" and discard the other "Verification started" message.

For now, I suggest we keep this simple message modification and here are my suggestions after building on yours:

- `Resolved (not verified)`
- `No errors` (might be a bit negative)
- `Resolution OK`

For the icons, here is my reflection with grades over 10 on how I like them

I don't like `gear` image (2/10) because it's invite the user to click on it for settings, which is not the affordance that we want.
`lightbulb` image (7/10) is better, although it's often used as an indication of an hint and invites clicking as well.

* `book` image (8/10) or `library` image (7/10), meaning Dafny can read the document, it's readable.
* `check` image (5/10) or `pass`  image (5/10), although I'm not sure here because it might indicate that everything is all right
* `feedback` image (6/10), `comment` image (6/10) or `output` image (7/10) (we give you good feedback on your code, but that's it, we don't provide any guarantees)
* `file-code` image (7/10), we recognize the code of the file.
* `pulse` image (4/10) (the file is in good shape but not all checked passed) (might be a bit unreadable and look like a M)
* `shield` image (5/10) or `wrench` (5/10) image (the file is proof-read against silly mistakes, with a technician's help, it's still awaiting ghost proofs)

If you're ok with it, I'd choose books.

camrein:
Thanks a lot for your effort. You've shared some good thoughts.

> In that case, we should replace the message "Compilation succeeded" by just the message "Verification started" and discard the other "Verification started" message.

That sounds to me that you're talking about the naming of the status messages in the language server. I propose we leave them as is to not break backward compatibility. Nevertheless, we're free to change the message displayed in VSCode.


> For now, I suggest we keep this simple message modification and here are my suggestions after building on yours:
> 
>     * `Resolved (not verified)`
> 
>     * `No errors` (might be a bit negative)
> 
>     * `Resolution OK`

As you mentioned, `no errors` sounds a bit negative. Moreover, it could lead people into thinking that there are no verification errors. Although, making the difference more clear is the primary goal of this PR. So for clarity, I'd opt for  `Resolved (not verified)`.

> For the icons, here is my reflection with grades over 10 on how I like them

I like your reasoning about the various icons and they make sense to me. After giving some additional thought, I guess `book` is a good option because it is not often used within VSCode. At least I haven't seen it often.

PR #93, opened 2021-12-02
Fixed stack overflow on the language server

MikaelMayer:
This fixes https://github.com/dafny-lang/dafny/issues/1213 by default for mac users, instead of having them run into this issue.


camrein:
Wasn't this fixed with https://github.com/dafny-lang/dafny/pull/1449? Sorry I've missed closing the issue you've referenced.
@keyboardDrummer, can you confirm that the mentioned PR resolved the issue?

MikaelMayer:
After several tests, this PR does not fix the problem that this Dafny user is encountering on a Mac using the latest version of the plugin (stuck on "Verifying..."), so I'll switch from draft to closed.

PR #104, opened 2021-12-16
feat(notifications): Display verification progress messages, if any

cpitclaudel:
This implements a simple mechanism to report plain-text details about the
verification process.

This is the companion PR of https://github.com/dafny-lang/dafny/pull/1655

### DIFF
--- a/src/language/api/compilationStatus.ts
+++ b/src/language/api/compilationStatus.ts
@@ -5,6 +5,7 @@ export enum CompilationStatus {
   ResolutionFailed = 'ResolutionFailed',
   CompilationSucceeded = 'CompilationSucceeded',
   VerificationStarted = 'VerificationStarted',
+  VerificationInProgress = 'VerificationInProgress',
   VerificationFailed = 'VerificationFailed',
   VerificationSucceeded = 'VerificationSucceeded'
 }
@@ -13,6 +14,7 @@ export interface ICompilationStatusParams {
   uri: DocumentUri;
   version?: integer;
   status: CompilationStatus;
+  message?: string | null;
 }
 
 export interface IVerificationStartedParams {
--- a/src/ui/compilationStatusView.ts
+++ b/src/ui/compilationStatusView.ts
@@ -17,10 +17,10 @@ function toStatusMessage(status: CompilationStatus, message?: string | null): st
   case CompilationStatus.CompilationSucceeded:
     return Messages.CompilationStatus.CompilationSucceeded;
   case CompilationStatus.VerificationStarted:
-    return Messages.CompilationStatus.Verifying;
+    return `${Messages.CompilationStatus.Verifying}...`;
   case CompilationStatus.VerificationInProgress:
     return message != null
-      ? `${Messages.CompilationStatus.VerificationInProgressHeader} ${message}`
+      ? `${Messages.CompilationStatus.Verifying} ${message}...`
       : Messages.CompilationStatus.Verifying;
   case CompilationStatus.VerificationSucceeded:
     return Messages.CompilationStatus.VerificationSucceeded;
--- a/src/ui/compilationStatusView.ts
+++ b/src/ui/compilationStatusView.ts
@@ -8,7 +8,7 @@ import { Messages } from './messages';
 
 const StatusBarPriority = 10;
 
-function toStatusMessage(status: CompilationStatus): string {
+function toStatusMessage(status: CompilationStatus, message?: string | null): string {
   switch(status) {
   case CompilationStatus.ParsingFailed:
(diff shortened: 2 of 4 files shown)


src/language/api/compilationStatus.ts
```diff
@@ -13,6 +14,7 @@ export interface ICompilationStatusParams {
   uri: DocumentUri;
   version?: integer;
   status: CompilationStatus;
+  message?: string; // FIXME this expects a string or undefined but it receives null
```
keyboardDrummer, line 17: Could you resolve the FIXME?

cpitclaudel: Yep, I need to figure how to do that ^^

src/ui/compilationStatusView.ts
```diff
@@ -18,6 +18,8 @@ function toStatusMessage(status: CompilationStatus): string {
     return Messages.CompilationStatus.CompilationSucceeded;
   case CompilationStatus.VerificationStarted:
     return Messages.CompilationStatus.Verifying;
+  case CompilationStatus.VerificationInProgress:
+    return message ?? Messages.CompilationStatus.Verifying;
```
MikaelMayer, line 22: The message will say something like this:

`Verifying CheckWellformed$$_module.__default.ToRelative ...`

which is arguably ugly.
Would you be able to remove the `CheckWellformed$$`, the `_module` if it is there (it's the default module) and  `__default` if it is there (it's the default class) ?

MikaelMayer (changes requested):

src/ui/compilationStatusView.ts
cpitclaudel: Will do, this way we'll be in line with Emacs too.  In this PR or on the Dafny side?

MikaelMayer: I think it's better to do it on the Dafny side, since else we make the IDE start to be aware of Boogie, which we would like to avoid.

cpitclaudel: Perfect, will update the pull request right away

src/language/api/compilationStatus.ts
cpitclaudel: Maybe @camrein can advise? 

Hi @camrein, if you have time: I have a problem with the the type of the `message?` field.  On the C# side the types is `string?`, and when serializing a `message` with value `null` the C# server produces `"message": null` in JSON (instead of leaving it undefined).  I've changed to `message?: string | null` on the JS side, which if fine but a bit ugly.  Can you think of something better?

Relatedly: I see that we have `version?: int` above; are we sure that the server omits the `version` field when `version` is `null`, or should that have been `version?: int | null` on the extension side?

camrein: > Hi @camrein, if you have time: I have a problem with the the type of the `message?` field.

The problem, as you might already think, is that C# doesn't know `undefined`.
To correctly represent this, you'd have to express this with a type within the language server. For example, create a sub-type of `CompilationStatusParams` such as `CompilationStatusWithMessageParams`.

> Relatedly: I see that we have `version?: int` above; are we sure that the server omits the `version` field when `version` is `null`, or should that have been `version?: int | null` on the extension side?

Not really. It'll be sent as `null` as well; thus, the correct type would be `version: int | null`. I copied this style from OmniSharp. However, this style is primarily used for client-to-server communication so it's not entirely correct. We could omit the `version` field since it's not used yet.

Side-note: As of now, apart from one or two exceptions, the code within the extension treats `undefined` and `null` the same.

cpitclaudel: Thanks for the clarification.  Given this it's probably fine to stick with what I did (make the type of `message?` be `string | null`)!

src/ui/compilationStatusView.ts
```diff
@@ -18,6 +18,10 @@ function toStatusMessage(status: CompilationStatus): string {
     return Messages.CompilationStatus.CompilationSucceeded;
   case CompilationStatus.VerificationStarted:
     return Messages.CompilationStatus.Verifying;
+  case CompilationStatus.VerificationInProgress:
```
camrein, line 21: Would you mind formatting it like the following and use string interpolation?
```ts
    return message != null
      ? `${Messages.CompilationStatus.VerificationInProgressHeader} ${message}`
      : Messages.CompilationStatus.Verifying;
```

You could also add the corresponding eslint rule to `.eslintrc.json`:
```json
"operator-linebreak": [ "warn", "before" ]
```

cpitclaudel: Done, thanks!

Issue #110, opened 2021-12-30
Possibly-overspecific supported runtimes pattern?


dijkstracula:
Dear Dafny team,

Apologies in advance if this is not in fact an issue but simply the state of the world.

I've been working on getting Dafny running on an M1 (arm64) Mac.  I've been following the [Dafny INSTALL wiki](https://github.com/dafny-lang/dafny/wiki/INSTALL) except that I've dropped in a custom-compiled Z3 4.8.5 since there does not appear to be an uploaded non-Intel release.  Luckily, the Dafny executable does appear to work on my machine (insofar as my trivial "hello world" program verifies, as does running `./Scripts/quicktest.sh`.

Now that the language itself appears to be running on this hardware, I'd next like to use the VSCode plugin with the language server - however, since no arm64 binaries for ASP .NET Core 5.0 have been released, I'm stuck using 6.0.  The Wiki does mention that Dafny itself runs under 6.0 (indeed, the wording suggests it's required); however, the [supported runtimes regex](https://github.com/dafny-lang/ide-vscode/blob/f087194118091e710f5f83882e0d4c4ed81cf6ef/src/constants.ts#L40) explicitly only matches against 5.0, so VSCode pops up a "There is no compatible dotnet runtime installed" dialog.

I'm not familiar with the .NET ecosystem, but I wonder whether this could safely be extended to support 5.0 as well as later versions?  Thanks!

camrein:
Thanks a lot for your suggestion.
As far as I remember, .NET (Core) does not guarantee backward compatibility. Unfortunately, I've only found [this](https://github.com/dotnet/runtime/issues/1959#issuecomment-576937893) GitHub issue confirming my statement.

I see that this check might be a bit too restrictive. Dafny 3.3 targets .NET Core 5.0, making sense for this compatibility check. However, it's hindering the next iteration, which targets 6.0, besides the issue you've already mentioned.

So there are a few things we have to keep in mind:

- The VSCode intends to support various Dafny versions in the future. Therefore, fixing the release to a specific .NET Version becomes problematic. For this case, we'd have to create a compatibility matrix.
- I've seen many people compile Dafny from source but not the VSCode extension. So the VSCode extension may not be aware of an upcoming Dafny release that may target another dotnet version.

Therefore, I propose that we either allow disabling the compatibility check or let people configure the targeted dotnet version themselves.
[...]

PR #111, opened 2021-12-31
checkDotnetInstallation: do not error out with ASP.NET Core Runtime 6.0

dijkstracula:
As discussed in #110, the check that the installed dotnet is precisely version 5.0 might be overly specific.  (In particular, the exact requirement for version 5.0 means no M1 Mac users can use the plugin since only x64 binaries are available for that version of the runtime).

This patch simply relaxes the regex that is matched against the output of `dotnet --list-runtimes` to allow either versions 5.0 or 6.0.

I forked the plugin and wrote this patch for my own personal use; however, I thought perhaps the team would like to consider whether it's worth merging in advance of Dafny 3.4, which @camrein mentioned will be targeting 6.0.  Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ You can find [examples below](#examples).
 
 ## Requirements
 
-The plugin requires at least .NET Core 5.0 (the ASP.NET Core 5.0 runtime to be more specific) to run the _Dafny Language Server_. Please download a distribution from [Microsoft](https://dotnet.microsoft.com/download).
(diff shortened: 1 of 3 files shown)


camrein:
Sorry for the late reply. Thanks a lot for sharing your changes.
In my opinion we can release an update that relaxes the requirement as suggested. However, I haven't tested the compatibility of the Dafny compiler and language server thoroughly. 
@keyboardDrummer @MikaelMayer You may have more experience in terms of the compatibility of Dafny 3.3 to .NET 6.0. Do you have any input?

keyboardDrummer (approved):

keyboardDrummer:
I've used .NET 6 to work with Dafny 3.3 and had no issues.

camrein:
Thanks for your feedback @keyboardDrummer.

PR #113, opened 2022-01-05
Fix empty configuration path that leads to the extension not working

### LINKED ISSUE #105: When using a custom Dafny.dll and language server reset is hard
If I experiment with a custom version of Dafny,
reseting the IDE back takes reseting the config values to default
AND uninstall/reinstall.

I would be nice if I could just reset it in the config

MikaelMayer:
If someone deletes a configured language server runtime or compiler path, there will remain an empty string.

Previously, it would have been interpreted wrongly like a "relative path", and users would have needed to click on the gears icon next to the setting, and click "reset", which is unintuitive, since it was empty when they first edited it.

With this commit, it is correctly interpreted as "switch back to default"

This would have saved me the first time, and I just learned it would have saved someone else from spending a few hours of trouble.

Fixes #105

### DIFF
--- a/src/language/dafnyInstallation.ts
+++ b/src/language/dafnyInstallation.ts
@@ -15,16 +15,9 @@ import Configuration from '../configuration';
 const ArchiveFileName = 'dafny.zip';
 const mkdirAsync = promisify(fs.mkdir);
 
-// If someone deletes a configured language server runtime or compiler path, there will remain an empty string.
-// Previously, it would have been interpreted wrongly like a "relative path".
-// Now it is interpreted as "switch back to default"
-function ifNullOrEmpty(a: string | null, b: string) {
-  return a === "" || a == null ? b : a;
-}
-
 export function getCompilerRuntimePath(context: ExtensionContext): string {
-  const configuredPath = ifNullOrEmpty(Configuration.get<string | null>(ConfigurationConstants.Compiler.RuntimePath)
-    , LanguageServerConstants.DefaultCompilerPath);
+  const configuredPath = Configuration.get<string | null>(ConfigurationConstants.Compiler.RuntimePath)
+    || LanguageServerConstants.DefaultCompilerPath;
   if(!path.isAbsolute(configuredPath)) {
     return path.join(context.extensionPath, configuredPath);
   }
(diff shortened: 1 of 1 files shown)


src/language/dafnyInstallation.ts
```diff
@@ -15,17 +15,24 @@ import Configuration from '../configuration';
 const ArchiveFileName = 'dafny.zip';
 const mkdirAsync = promisify(fs.mkdir);
 
+// If someone deletes a configured language server runtime or compiler path, there will remain an empty string.
+// Previously, it would have been interpreted wrongly like a "relative path".
+// Now it is interpreted as "switch back to default"
+function ifNullOrEmpty(a: string | null, b: string) {
+  return a === "" || a == null ? b : a;
```
fabiomadge, line 22:
```suggestion
  return a === '' || a == null ? b : a;
```

seebees:
I suspect that this will also fix #105.

src/language/dafnyInstallation.ts
```diff
@@ -15,17 +15,24 @@ import Configuration from '../configuration';
 const ArchiveFileName = 'dafny.zip';
 const mkdirAsync = promisify(fs.mkdir);
 
+// If someone deletes a configured language server runtime or compiler path, there will remain an empty string.
+// Previously, it would have been interpreted wrongly like a "relative path".
+// Now it is interpreted as "switch back to default"
+function ifNullOrEmpty(a: string | null, b: string) {
```
keyboardDrummer, line 21: This seems like a strange function because `null` and `''` are both falsy in JavaScript, so instead of `isNullOrEmpty(a,b)`, I think you could also do `a || b`

MikaelMayer: Good point. I'll even inline a || b since it's so simple.

keyboardDrummer (approved):
There's a new warning now: "Unexpected nullable string value in conditional. Please handle the nullish/empty cases explicitly". 

I guess eslint disagrees with my suggestion. Feel free to use the previous version you had since that it was it's asking you to do, or otherwise turn off the warning with `// eslint-disable-next-line`

MikaelMayer:
> // eslint-disable-next-line

I reverted to the previous one, but based on your suggestion, I simplified the conditional with `a ? a : b` instead of `a == ='' || a == null ? b : a`

camrein (approved):
> > // eslint-disable-next-line
> 
> I reverted to the previous one, but based on your suggestion, I simplified the conditional with `a ? a : b` instead of `a ==='' || a == null ? b : a`

I'm fine with both solutions. I tend to configure linters/static analyzers extremely strict and relax rules where needed. I suggest leaving the rule as is because the situation we've got here is rather seldom in my opinion. But an exception for this line is absolutely justified.

keyboardDrummer (approved):

PR #116, opened 2022-01-14
Added option to install a preferred dafny version

### LINKED ISSUE #107: Versioned installation paths
Currently the VSCode Dafny IDE extension installs a specific version of Dafny.  To use a different version, users have to install Dafny manually and point the extension to that.  It would be nice to make this easier.

I've recently implemented auto-installation for Dafny in Emacs (I'm porting the Emacs mode to use LSP), and in that code I used the following design:

* There is a user setting to specify which version of Dafny to use.  The plugin has a hardcoded list of known versions that it is compatible with, and the value of that setting must be one of these.  It defaults to the highest known version.
* The Dafny binary that the IDE uses is either the one the user specified manually, if any, or the one corresponding to the "version" user setting above.
* If the binary corresponding to the selected version is not installed, the IDE downloads that version of Dafny and saves it to a version-dependent path, which makes it easy to check whether a given version of Dafny is installed.

I experimented with this in Emacs first since changing the VSCode extension wasn't trivial (given that we already had many live installs).  But since we are in fact publishing a new Dafny extension under `dafny-lang`, it might be worth making a similar change to the Dafny VSCode extension, to make it easy for users to switch versions (just change the version and restart the server or IDE).

If we decide to do this, it would be best to do it soon; this way we wouldn't have to worry about backwards compatibility, since we're publishing a new plugin.

camrein:
Resolves #107 

I've intentionally left out the following versions:

- 3.1.0: Requires the creation of an empty `nlog.config` file by the extension
- 3.0.0: The language server and compiler are provided by two separate archives

I'd love some thoughts about some unclean changes. The `constants.ts` file only served actual constants until now. I've changed some constants to functions to better serve the configuration need. The way it's now makes it easier to adapt these strings in the future, but they're now functions.

Moreover, what do you think about showing the "outdated installation" messages for any version selection prior to the one specified by `LatestVersion`? That's what's happening now.


src/language/dafnyInstallation.ts
```diff
@@ -65,14 +81,14 @@ export class DafnyInstaller {
...
     if(version === LanguageServerConstants.UnknownVersion) {
       return true;
     }
     const [ givenMajor, givenMinor ] = version.split('.');
-    const [ requiredMajor, requiredMinor ] = LanguageServerConstants.RequiredVersion.split('.');
-    return givenMajor > requiredMajor
-      || givenMajor === requiredMajor && givenMinor >= requiredMinor;
+    const [ latestRequired, latestMinor ] = LanguageServerConstants.LatestVersion.split('.');
+    return givenMajor > latestRequired
+      || givenMajor === latestRequired && givenMinor >= latestMinor;
```
cpitclaudel, line 91: Shouldn't this have three components?

cpitclaudel:
Looks great!  I sent you the notes I had about this feature :)

src/language/dafnyInstallation.ts
cpitclaudel: Something like this:

```
  public static isMinimumRequiredLanguageServer(version: string): boolean {
      const v = version.split('.');
      const req = LanguageServerConstants.MinVersion.split('.');
      for (var i = 0; i < req.length; i++) {
          var vi = v[i] ?? "0";
          var ri = req[i];
          if (vi == ri)
              continue
          return vi > ri;
      }
      return true;
  }
```

cpitclaudel:
> I've changed some constants to functions to better serve the configuration need. The way it's now makes it easier to adapt these strings in the future, but they're now functions.

I think this is completely fine.  Alternatively you could keep the constant parts in that file and move the functions to the installation path, but I think things are fine as-is.

src/language/dafnyInstallation.ts
```diff
@@ -54,7 +70,7 @@ function getDafnyPlatformSuffix(): string {
 
 function getDafnyDownloadAddress(): string {
   const baseUri = LanguageServerConstants.DownloadBaseUri;
-  const version = LanguageServerConstants.RequiredVersion;
+  const version = getConfiguredVersion();
```
cpitclaudel, line 73: I would have decoupled the setting from this file and passed the version to the constructor of this class instead, but it's a stylistic issue.

src/language/dafnyInstallation.ts
camrein: You're right, it's probably best to include the patch part of the version string too. I can't really remember the reasoning behind excluding it anymore, or it might no longer apply.

src/language/dafnyInstallation.ts
camrein: Most of the version handling is rather self-contained by now, i.e., the only version that is present outside this module is the one received during startup. The configured and installed version is only present and used within this module. Therefore, I think we can leave it as-is for the moment.
However, I've got the feeling that the class/module got a bit messy. I'll look if there are some reasonable refactorings.

Last but not least, I want to give automated integration testing another try. There were some code changes that were easy to miss. The following tests would be desirable here:

- `latest` installs and launches the version specified by `LatestVersion`
- Any of the *supported* versions can be installed and launched

camrein:
@cpitclaudel I've updated the version check to compare against all the available version parts and added some output to the Dafny console.
Unfortunately, I don't think that I'll have the capacity to implement the automatic testing I'd like. But I think this feature might be useful when introducing Dafny 3.4.

src/language/dafnyInstallation.ts
```diff
@@ -65,14 +81,28 @@ export class DafnyInstaller {
...
+      if(given < latest) {
+        this.writeStatus(`the installed Dafny version is older than the latest: ${version} < ${latestVersion}`);
+        return false;
+      }
+      if(given > latest) {
+        this.writeStatus(`the installed Dafny version is newer than the latest: ${version} > ${latestVersion}`);
+        return true;
+      }
+    }
+    this.writeStatus(`the installed Dafny version is the latest known: ${version} = ${latestVersion}`);
```
cpitclaudel, line 104: This will report `=` for `3.2` vs `3.2.1`, won't it?

cpitclaudel (approved):
Great work, thanks!

src/language/dafnyInstallation.ts
camrein: Exactly. I've taken the minimum of both lengths to account for cases where the full version string is not available for one or another.
Moreover, it's relevant to account for the build-version that is provided by the language server:
```
the installed Dafny version is the latest known: 3.3.0.31104 = 3.3.0
```

PR #122, opened 2022-01-25
feat: Dafny Plugins support

MikaelMayer:
This PR requires https://github.com/dafny-lang/dafny/pull/1817 to be merged so that it works properly

This PR features a new option "Dafny Plugins", which essentially is a list of comma-separated DLLs that Dafny can use in various places to augment its pipeline. It also supports arguments, as explained here:
https://github.com/dafny-lang/dafny/tree/master/Source/DafnyLanguageServer#plugins

Also, I added logging information about what launch arguments were used to start the DafnyLanguageServer.


cpitclaudel:
I think this option should be a list instead of a comma-separated string

MikaelMayer:
> I think this option should be a list instead of a comma-separated string

Unless you know something I'm not aware of, lists are not at all user friendly in VSCode. With a list, you get the interface:
image
which provides no clue on how to add this setting to settings.json (look how people would have to guess it's `dafny.compilerArgs` with this case)

If you make it a string, VSCode makes it easy to edit and use.

cpitclaudel:
> Unless you know something I'm not aware of, lists are not at all user friendly in VSCode. 

https://code.visualstudio.com/updates/v1_37#_settings-editor-support-for-string-arrays

Maybe it's because you used an optional list instead of a list?

camrein:
Thanks for linking that doc @cpitclaudel. I think the reason that we're not seeing this kind of UI interaction is that we do not specify the type of the items. Therefore, the `dafny.compilerArgs` option should look like the following:
```cs

        "dafny.compilerArgs": {
          "type": "array",
          "items": {
            "type": "string"
          },
          "default": [
            "/verifyAllModules",
            "/compile:1",
            "/spillTargetCode:1"
          ],
          "description": "Optional array of strings as Dafny compilation arguments"
        }
```

MikaelMayer:
> > Unless you know something I'm not aware of, lists are not at all user friendly in VSCode.
> 
> https://code.visualstudio.com/updates/v1_37#_settings-editor-support-for-string-arrays
> 
> Maybe it's because you used an optional list instead of a list?

Fantastic ! That's because the type of items was not specified. I'm going to do a separate PR so that it becomes the default.

image

Here is the PR. I'll update this PR accordingly.
https://github.com/dafny-lang/ide-vscode/pull/124

src/language/dafnyLanguageClient.ts
```diff
@@ -51,6 +52,14 @@ function getMarkGhostStatementsArgument(): string {
   return `--ghost:markStatements=${Configuration.get<string>(ConfigurationConstants.LanguageServer.MarkGhostStatements)}`;
 }
 
+function getDafnyPluginsArgument(): string {
+  let plugins = Configuration.get<string[]>(ConfigurationConstants.LanguageServer.DafnyPlugins);
+  if(plugins === null) {
+    return '';
+  }
+  return `--dafny:plugins=${plugins.join(',')}`;
```
cpitclaudel, line 60: But on the Dafny side, don't we want to pass multiple copies of the flag instead of using a `,`?

cpitclaudel (changes requested):
We need to wait for the Dafny side before we merge

src/language/dafnyLanguageClient.ts
MikaelMayer: Indeed, I forgot to push my changes for this VSCode plugin

camrein (changes requested):
There are some linter warnings that should be resolved (see the results of the GitHub action). If you're using VSCode, install the ESLint extension.
You may also run the linter manually from CLI using `npm run lint`.

package.json
```diff
@@ -155,6 +155,14 @@
           "default": "true",
           "description": "Mark ghost statements in the code (requires restart and Dafny 3.4.0+)"
         },
+        "dafny.dafnyPlugins": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "default": [],
+          "description": "Absolute paths to assemblies to be used as plugins (requires restart and Dafny 3.4.0+).\nExample 1: /user/home/dafnyplugin.dll\nExample 2: /user/home/dafnyplugin.dll,oneArgument\nExample 3: /user/home/dafnyplugin.dll,\"argument with space and \\\" escaped quote\" secondArgument"
```
cpitclaudel, line 164: What happens if there's a comma in a plugin argument?  (Do we really need plugin arguments?)

src/language/dafnyLanguageClient.ts
```diff
@@ -51,6 +52,21 @@ function getMarkGhostStatementsArgument(): string {
...
+  let plugins = Configuration.get<string[]>(ConfigurationConstants.LanguageServer.DafnyPlugins);
+  if(plugins === null) {
+    return '';
+  }
+  plugins = plugins.filter(x => x !== null && x !== '');
+  const result = [];
+  for(const i in plugins) {
+    const possiblyEscaped = plugins[i].includes(' ') && !plugins[i].startsWith('"')
+      ? '"' + plugins[i].replace(/\\/g, '\\\\').replace(/"/g, '\\"') + '"' : plugins[i];
+    result.push(`--dafny:plugins:${i}=${possiblyEscaped}`);
```
cpitclaudel, line 65: This is dangerous.  I have double-checked the escaping, but it gets injected right into another command line, which is scary.

src/language/dafnyLanguageClient.ts
```diff
@@ -51,6 +52,21 @@ function getMarkGhostStatementsArgument(): string {
...
+  }
+  return result.join(' ');
+}
+
```
cpitclaudel, line 69: If there are command line arguments, they should be escaped before being concatenated, or better yet we should use an API that takes a list of arguments, not a string.

cpitclaudel: And in fact this is what we are doing, so this function should return a `string[]`, not a string (and in the caller you can use `...getDafnyPluginsArguments()`

src/language/dafnyLanguageClient.ts
```diff
@@ -61,9 +77,10 @@ export class DafnyLanguageClient extends LanguageClient {
     return this.sendRequest<ICounterExampleItem[]>('dafny/counterExample', param);
   }
 
-  public static async create(context: ExtensionContext): Promise<DafnyLanguageClient> {
+  public static async create(context: ExtensionContext, statusOutput: OutputChannel): Promise<DafnyLanguageClient> {
     const dotnetExecutable = await getDotnetExecutablePath();
     const launchArguments = [ getLanguageServerRuntimePath(context), ...getLanguageServerLaunchArgs() ];
+    statusOutput.appendLine(`Language server arguments: ${launchArguments.join(' ')}`);
```
cpitclaudel, line 83: This needs escaping

cpitclaudel (changes requested):
Looks OK, but we shouldn't concatenate command line arguments

src/language/dafnyLanguageClient.ts
MikaelMayer: I removed this code, I now leave it entirely to the users to actually provide correct escape, which in most cases is not needed, and if needed, it matches the command-line escape that Dafny itself will unescape, not the plugin.

package.json
MikaelMayer: Only the first comma delimits the plugin from the argument list.
Yes we need plugin arguments, so that the same plugin can be configured through the class PluginConfiguration.
There are several use case for plugin arguments
* If the plugin provides several rewriters, to activate one of them only
* If rewriters have different options, such as "experimental" options, it makes it possible to activate them by changing a command-line flag.

MikaelMayer:
I just push some changes but realized we need to simplify how the parsing of argument works at the Dafny level.

MikaelMayer:
I just tested it again on one of the Dafny tests, it works like a charm.

src/language/dafnyLanguageClient.ts
```diff
@@ -51,6 +52,20 @@ function getMarkGhostStatementsArgument(): string {
...
 
+function getDafnyPluginsArgument(): string[] {
+  const plugins = Configuration.get<string[]>(ConfigurationConstants.LanguageServer.DafnyPlugins);
+  if(plugins === null || !Array.isArray(plugins)) {
+    return [];
+  }
+  const result = [];
+  for(const i in plugins) {
+    const plugin = plugins[i];
+    if(plugin === null || plugin === '') continue;
```
camrein, line 63: Please always use curly braces.

You may add the following rule to `.eslintrc.json`:
```json
    "curly": "warn",
```

src/language/dafnyLanguageClient.ts
```diff
@@ -61,9 +76,14 @@ export class DafnyLanguageClient extends LanguageClient {
     return this.sendRequest<ICounterExampleItem[]>('dafny/counterExample', param);
   }
 
-  public static async create(context: ExtensionContext): Promise<DafnyLanguageClient> {
+  public static async create(context: ExtensionContext, statusOutput: OutputChannel): Promise<DafnyLanguageClient> {
     const dotnetExecutable = await getDotnetExecutablePath();
     const launchArguments = [ getLanguageServerRuntimePath(context), ...getLanguageServerLaunchArgs() ];
+    statusOutput.appendLine(`Language server arguments: ${launchArguments.map(oneArgument =>
```
camrein, line 82: Please extract the "prettifying" of the arguments into a separate function for readability. The best would be to extract the whole part that's responsible to turn the `launchArguments` array into a string.

src/language/dafnyLanguageClient.ts
```diff
@@ -51,6 +52,20 @@ function getMarkGhostStatementsArgument(): string {
   return `--ghost:markStatements=${Configuration.get<string>(ConfigurationConstants.LanguageServer.MarkGhostStatements)}`;
 }
 
+function getDafnyPluginsArgument(): string[] {
+  const plugins = Configuration.get<string[]>(ConfigurationConstants.LanguageServer.DafnyPlugins);
+  if(plugins === null || !Array.isArray(plugins)) {
+    return [];
+  }
+  const result = [];
+  for(const i in plugins) {
```
camrein, line 61: It should generally be avoided to use `for(... in ...)` to iterate over the indices of an array. For example, the order of the keys (which are the indices for arrays) is not consistent:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#array_iteration_and_for...in

Threfore, prefer using a conventional for loop to iterate over the indices.

However, I think this whole for-loop can be neatly refactored into a map and filter call. Something like this:
```js
return plugins
  .filter(plugin => plugin != null && plugin !== '')
  .map((plugin, index) => `--dafny:plugins:${index}=${plugin}`);
```

camrein (changes requested):

src/language/dafnyLanguageClient.ts
MikaelMayer: Well, I prefer comprehensions too. I also added parentheses around your expression because the semicolon is optional in JavaScript.

camrein: > Well, I prefer comprehensions too. I also added parentheses around your expression because the semicolon is optional in JavaScript.

Because the semicolons are optional in JavaScript/TypeScript, I configured eslint to enforce them. It's way to easy to forget about some implicit semicolon, especially when context switching between multiple languages.

src/language/dafnyLanguageClient.ts
```diff
@@ -61,9 +74,19 @@ export class DafnyLanguageClient extends LanguageClient {
...
+        ? '"' + oneArgument.replace(/\\/g, '\\\\').replace(/"/g, '\\"') + '"'
+        : oneArgument
+    ).join(' ');
+  }
+
+  public static async create(context: ExtensionContext, statusOutput: OutputChannel): Promise<DafnyLanguageClient> {
     const dotnetExecutable = await getDotnetExecutablePath();
     const launchArguments = [ getLanguageServerRuntimePath(context), ...getLanguageServerLaunchArgs() ];
+    statusOutput.appendLine(
+      `Language server arguments: ${DafnyLanguageClient.argumentsToCommandLine(launchArguments)}`);
```
camrein, line 89: Honestly, I don't like these random line breaks IntelliJ/Rider/WebStorm does, it should at least move the closing parenthesis to a new line too. Unfortunately, I haven't found an eslint rule to enforce this yet.

PR #125, opened 2022-01-31
feat: Make status bar clickable and open a Dafny command menu

cpitclaudel:
We've had users ask for a menu to restart Dafny. Here is how we are going to do it.

When users click on this:
image
They will be shown the following command menu:
image
The first item is omitted if the status is not `Verification Failed` or `Verifying`

Note: It seems that calling `await` on `client.stop` before restarting is not enough: I get the following stack trace, but it does not block the restart per se.

```
rejected promise not handled within 1 second: Error [ERR_STREAM_WRITE_AFTER_END]: write after end
c:\Users\cpitcla\AppData\Local\Programs\Microsoft VS Code\resources\app\out\bootstrap-fork.js:5
stack trace: Error [ERR_STREAM_WRITE_AFTER_END]: write after end
	at Socket.write (internal/streams/writable.js:292:11)
	at c:\Users\cpitcla\git\dafny\ide-vscode\dist\extension.js:6992:29
	at new Promise (<anonymous>)
	at WritableStreamWrapper.write (c:\Users\cpitcla\git\dafny\ide-vscode\dist\extension.js:6982:16)
	at StreamMessageWriter.doWrite (c:\Users\cpitcla\git\dafny\ide-vscode\dist\extension.js:8233:33)
	at c:\Users\cpitcla\git\dafny\ide-vscode\dist\extension.js:8224:29
	at runNextTicks (internal/process/task_queues.js:58:5)
	at processImmediate (internal/timers.js:434:9)
	at process.callbackTrampoline (internal/async_hooks.js:131:14)
c:\Users\cpitcla\AppData\Local\Programs\Microsoft VS Code\resources\app\out\bootstrap-fork.js:5
rejected promise not handled within 1 second: Error [ERR_STREAM_WRITE_AFTER_END]: write after end
c:\Users\cpitcla\AppData\Local\Programs\Microsoft VS Code\resources\app\out\bootstrap-fork.js:5
stack trace: Error [ERR_STREAM_WRITE_AFTER_END]: write after end
	at Socket.write (internal/streams/writable.js:292:11)
	at c:\Users\cpitcla\git\dafny\ide-vscode\dist\extension.js:6992:29
	at new Promise (<anonymous>)
	at WritableStreamWrapper.write (c:\Users\cpitcla\git\dafny\ide-vscode\dist\extension.js:6982:16)
	at StreamMessageWriter.doWrite (c:\Users\cpitcla\git\dafny\ide-vscode\dist\extension.js:8233:33)
	at c:\Users\cpitcla\git\dafny\ide-vscode\dist\extension.js:8224:29
	at runNextTicks (internal/process/task_queues.js:58:5)
	at processImmediate (internal/timers.js:434:9)
	at process.callbackTrampoline (internal/async_hooks.js:131:14)
```


camrein:
I had to do some digging around where this was initially implemented:
https://github.com/dafny-lang/ide-vscode/blob/29bc7a56122e8b23993ab2035fa930b391d08650/src/dafnyLanguageServerStartup/serverInitializer.ts#L101-L110

I'd suggest to make it a command in the context (right click) menu again. I think people will have a hard time to figure out that they can interact with the status bar, but only when it's in a certain state. Especially when the action is not related inherently.

I'll try to have a look at your implementation by tomorrow.

cpitclaudel:
> I think people will have a hard time to figure out that they can interact with the status bar, but only when it's in a certain state. 

The code I wrote makes that action available as a regular command, but also makes the status bar (always) a shortcut to enumerate available commands.

> I'll try to have a look at your implementation by tomorrow.

Thanks!

camrein (commented):
The code looks good so far. Unfortunately, I couldn't track down the problem you experience yet.

About the question where the command should go. I have the feeling that it might better fit to `extension.ts` since the command mostly interacts with this code. However, it may make sense to extract this code into a new module.

src/extension.ts
```diff
@@ -28,6 +28,10 @@ export async function deactivate(): Promise<void> {
   await extensionRuntime?.dispose();
 }
 
+export async function restartServer(): Promise<void> {
```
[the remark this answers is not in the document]
MikaelMayer: I don't think it does introduce a circular dependency. This is a top-level function that invokes a class function.

src/extension.ts
```diff
@@ -99,4 +102,24 @@ class ExtensionRuntime {
   public async dispose(): Promise<void> {
     await this.client?.stop();
   }
+
+  public async startClientAndWaitForVersion() {
+    this.client = this.client ?? await DafnyLanguageClient.create(this.context, this.statusOutput);
```
cpitclaudel, line 107: Since we recycle this.client if it exists here, shouldn't we also exit right away if we already have a client?

src/extension.ts
```diff
@@ -99,4 +102,24 @@ class ExtensionRuntime {
...
+  }
+
+  public async restart(): Promise<void> {
+    this.statusOutput.appendLine('Terminating Dafny...');
+    await this.dispose();
+    // The first subscription is the statusOutput and should not be disposed.
+    for(let i = 1; i < this.context.subscriptions.length; i++) {
+      this.context.subscriptions[i].dispose();
+    }
+    this.context.subscriptions.splice(1, this.context.subscriptions.length);
```
cpitclaudel, line 120:
```suggestion
    this.context.subscriptions.splice(1);
```

(the second parameter is how many elements to delete), so it should be `this.context.subscriptions.length - 1` if you want to keep it, I think)

src/ui/statusBarActionView.ts
```diff
@@ -0,0 +1,71 @@
...
+
+Expected behavior
+-----------------
+
+
+Actual behavior
+---------------
+
+
+`.trimStart());
```
cpitclaudel, line 30: We could create a template in the repo for this?

src/ui/statusBarActionView.ts
```diff
@@ -0,0 +1,71 @@
...
+interface StatusBarAction {
+  label: string;
+  description: string;
+  command?: string;
+  uri?: string;
+}
+
+const ReportVerificationIssue: StatusBarAction = {
+  label: 'Report verification issue',
+  description: 'File an issue on Github about a verification problem you have.',
```
cpitclaudel, line 41: I would be careful with this; until now we have been pointing people to the "discussions" if they have trouble writing a proof

src/ui/statusBarActionView.ts
```diff
@@ -0,0 +1,71 @@
...
+  uri: fileIssueURL
+};
+
+
+const RestartDafny: StatusBarAction = {
+  label: 'Restart Dafny',
+  description: 'Relaunches the Dafny Language Server and reverify the current file.',
```
cpitclaudel, line 48:
```suggestion
  description: 'Relaunch the Dafny Language Server and reverify the current file.',
```

(or `reverifies`)

src/extension.ts
```diff
@@ -99,4 +102,24 @@ class ExtensionRuntime {
...
+    this.client = this.client ?? await DafnyLanguageClient.create(this.context, this.statusOutput);
+    this.client.start();
+    await this.client.onReady();
+    this.languageServerVersion = await this.getLanguageServerVersionAfterStartup();
+  }
+
+  public async restart(): Promise<void> {
+    this.statusOutput.appendLine('Terminating Dafny...');
+    await this.dispose();
+    // The first subscription is the statusOutput and should not be disposed.
```
cpitclaudel, line 116: Could we delete that one and recreate it as well?

src/extension.ts
MikaelMayer: I tested this hypothesis and restarting the server no longer works.
I think we still need to start the client - which will start the language server.

src/ui/statusBarActionView.ts
MikaelMayer: Good point. To make it clear why it's better to generate it from VSCode, I also included the version of Dafny and the version of the VSCode extension in the last commit.

src/ui/statusBarActionView.ts
MikaelMayer: I changed is so now the first command links to "discussions" if they need help for a proof, if verification is not successful.

src/extension.ts
MikaelMayer: I did not find out how to do it properly. Activate() is called only once.

src/ui/statusBarActionView.ts
```diff
@@ -0,0 +1,111 @@
...
+
+const ReportVerificationIssue = async (dafnyVersion: string, context: ExtensionContext): Promise<StatusBarAction> => ({
+  label: 'Report verification issue',
+  description: 'If Dafny is not behaving as expected',
+  uri: await fileIssueURL(dafnyVersion, context)
+});
+
+const RestartDafny: StatusBarAction = {
+  label: 'Restart Dafny',
+  description: 'IF the Dafny verifier seems stuck',
```
cpitclaudel, line 62:
```suggestion
  description: 'If Dafny seems unresponsive',
```

src/extension.ts
```diff
@@ -99,4 +102,27 @@ class ExtensionRuntime {
...
+    for(let i = 1; i < this.context.subscriptions.length; i++) {
+      this.context.subscriptions[i].dispose();
```
keyboardDrummer, line 118: Why do we need to do this? Shouldn't we only restart the LSP server?

keyboardDrummer: How about calling `client.stop()`, which tries to shutdown the language server and then disconnects from it, following by `client.start()` ?

MikaelMayer: `this.dispose()` actually calls `client.stop()`
`this.initializeClient()` actually calls `client.start();`, but with proper logging.
So I also think it's the right way of doing.

keyboardDrummer: But.. right now I feel that we're cleaning up too. We're removing all traces of the Dafny extension and then adding them again, right? Doing `client.stop()` followed by `client.start()` would I think just restart the LSP server.

keyboardDrummer: Do you know why it's not working? Is that a bug in `client.stop()` then? The code is here: https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/client.ts#L1231

Which should lead to the client sending a shutdown request to the server here: https://github.com/microsoft/vscode-languageserver-node/blob/536cc154323d4d5d66c225a9ad5c5f259057f6f9/client/src/common/client.ts#L2069

MikaelMayer: Thanks for sharing this code, I did not know where to find it exactly.

So it looks like I'm right: subscriptions are not part of the client, but of the context of the extension; however, many of them rely on the connection like this: https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/client.ts#L738

This `connection` is a snapshot that is [reset when the client restarts](https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/client.ts#L1444)
onNotification. Thus, it justifies that we need to dispose old notification disposables and recreate somes, otherwise, the extension would not work.

keyboardDrummer: Makes sense, thanks for the explanation!

keyboardDrummer (approved):

Issue #85, opened 2021-11-17
String highlighting in the IDE only


seebees:
The following in VSCode looks strange, however here in GitHub, it looks fine...
```dafny
method Foo() returns (res: Result<string, string>)
  {
    :- Need(true, "Not a string, the parens are doing it ().");
    :- Need(!false, "The code is higlites as a string and this string is code");
    :- Need(false == false, "???");
  }
```

camrein:
Thanks a lot for your report. The strings are indeed rendered strangely:
![image](https://user-images.githubusercontent.com/18049310/142372692-d4cebb5b-3624-4dbd-8b3c-523f2d108ee7.png)

I will check how to improve upon this. One thing I'll change is that strings are no longer highlighted over multiple lines since Dafny does not support multiline strings with `"`. This will avoid breaking the syntax highlighting for the whole file in case there's another issue.

PR #127, opened 2022-02-07
fix: Generic highlights in types and in method call

### LINKED ISSUE #68: Syntax highlighting of type parameters
When a type (like `MyClass` below) takes type parameters, the first type argument is syntax-highlighted in the expected way, but the remaining arguments are not. For example,

c

shows the first `bool` in blue, but the next `int` and `bool` in white. I would expect built-in types like `int` and `bool` to be colored the same way wherever they occur.

### LINKED ISSUE #79: Syntax highlighting of generic functions
```dafny
// `Foo` is a function, and `string` is a type
function Foo(bar: string) {}

// Foo is not a function, and `T` is not a type 
function Foo<T>(bar: T) {}
```

I suspect that is related to #68

MikaelMayer:
* Fixes #68 
Before:
image
After:
image

* Fixes #79 
Before:
image
After:
image

### DIFF
--- a/syntaxes/Dafny.tmLanguage
+++ b/syntaxes/Dafny.tmLanguage
@@ -54,7 +54,7 @@
 		</dict>
 		<dict>
 			<key>begin</key>
-			<string>(\w+)\s*\(</string>
+			<string>(\w+)\s*(&lt;[^&gt;]+&gt;)?\s*\(</string>
 			<key>beginCaptures</key>
 			<dict>
 				<key>1</key>
--- a/syntaxes/Dafny.tmLanguage
+++ b/syntaxes/Dafny.tmLanguage
@@ -308,19 +308,18 @@
 			<array>
 				<dict>
 					<key>begin</key>
-					<string>&lt;(\w+)</string>
-					<key>beginCaptures</key>
-					<dict>
-						<key>1</key>
-						<dict>
-							<key>name</key>
-							<string>storage.type.dafny</string>
-						</dict>
-					</dict>
+					<string>&lt;(?=\w+)</string>
 					<key>end</key>
 					<string>&gt;</string>
 					<key>name</key>
 					<string>meta.type.identifier.dafny</string>
+					<key>patterns</key>
+					<array>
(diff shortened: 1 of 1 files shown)


cpitclaudel (approved):

syntaxes/Dafny.tmLanguage
```diff
@@ -54,7 +54,7 @@
 		</dict>
 		<dict>
 			<key>begin</key>
-			<string>(\w+)\s*\(</string>
+			<string>(\w+)\s*(&lt;[^&gt;]+&gt;)?\s*\(</string>
```
cpitclaudel, line 57: Can you do this for signatures only?  I don't think it should happen in method bodies.

cpitclaudel: Because in a method body it doesn't work consistently, e.g. `Foo<Bar<Baz>>(xyz)`

keyboardDrummer:
Is there a way to write tests for our tmLanguage file ? Seems like it's done here: https://github.com/microsoft/TypeScript-TmLanguage/tree/master/tests

camrein:
+1 to the automized tests. We had a past issue with strings, that's why I've removed multi-line string highlighting from normal strings, see #85.

syntaxes/Dafny.tmLanguage
MikaelMayer: I fixed it for depths until 3.

MikaelMayer:
@camrein could you have a look again? I updated the description, included a test file for manually reviewing regression tests for now, and fixed one more issue. I just want someone else to have a look at it before merging it, even if I have the rights.

MikaelMayer:
I'm going to push more fixes to this branch, because I was able to break the tmLanguage code and find how everything worked together.

MikaelMayer:
> Outstanding work @MikaelMayer! Ready to release the new VSCode extension today.

Thanks ! I'm actually terrified with this grammar but I feel better prepared now to handle other syntax highlighting feature requests !

PR #129, opened 2022-02-09
chore: Prepare v2.2.0

camrein:
Since Dafny 3.4.0 is out, I'm preparing the release of a new VSCode extension to target said Dafny version.

PS: We may want to include #128 and #127 with this release too.
PPS: I think it makes sense now to switch to squash commits when merging a PR.


CHANGELOG.md
```diff
@@ -3,7 +3,12 @@
 ## 2.2.0
 
 - Added option to specify the preferred dafny version for automatic installation
+- Empty compiler/language server paths are now treated as unconfigured (use automatic installation)
+- Now accepting .NET 6.0 as a supported platform
```
keyboardDrummer, line 7: Nitpick I'd phrase this and similar messages as `Accept .NET 6.0 as a supported platform`, so you don't have to write every change in past tense.

keyboardDrummer (approved):

MikaelMayer:
> Do we want to include #122 with this release as well? In my opinion we should leave it for the next release since it's not supported by any of the publicly available Dafny versions.

Agreed.

Issue #132, opened 2022-02-14
Wrong syntax highlighting of set comprehensions


MikaelMayer:
With the new syntax highlighting, it wrongly believes that set comprehensions's first term are types (probably because of type ascription)

![image](https://user-images.githubusercontent.com/3601079/153892612-6377e3d4-7d19-4158-9ca2-4117e7b72480.png)

The second "id" is blue.

PR #133, opened 2022-02-14
fix: type ascription should not be applicable to double colon

MikaelMayer:
Note that the second "id" is no longer understood as a type (previously it was matching ": id" as a type ascription), but now it's understood a function correctly.

Before:
image

After:
image


MikaelMayer:
> Would you mind adding this change to the `regression-highlight.dfy` you introduced recently?

Excellent suggestion. I don't know why I didn't think of that in the first place, but yes that file was made for exactly that purpose ! Added.

camrein:
No worries, that happens from time to time. Thanks for adding it, as well as the `regression-highlight.dfy` in general, I love it!

resolves #132

Issue #139, opened 2022-02-23
VSCode sometimes not displaying red underlines for errors


cpitclaudel:
Copied from https://github.com/dafny-lang/dafny/issues/1854 by @emina :

> I've observed that the Dafny VSCode plugin sometimes fails to localize parsing and type errors by underlining them with the red squiggles.
> 
> This behavior is not easy to reproduce, but I'm reporting it here in case others run into it.
> 
> In general, it happens when I'm working on a large file (300+ LOC), and VSCode has been open for a long time (several days).
> 
> Even when the errors aren't localized in the UI, the "Problems" window does list the locations correctly, and they're also listed / found by running the verifier manually.

camrein:
Thanks for filing the GH issue. It's definitely strange that VSCode fails to create the squiggles, especially since VSCode receives the diagnostics since it's displaying it in the problems window. What I'm wondering if it somehow conflicts with the ghost highlighting feature?

cpitclaudel:
Fixed by #141.

Issue #126, opened 2022-01-31
Suggestions for ghost highlighting


RustanLeino:
I love the new ghost-highlighting feature in the Dafny extension. This feature adds a gray background for statements that are ghost. I have two further suggestions/wishes for this mode.

* I have found that `lemma` bodies are all grayed out. My first thought was that this is due lemmas always being ghost. However, in writing up this Issue, I'm finding that I can't always reproduce the situation. I have now written several `lemma` or `least lemma` declarations without their bodies being turned into gray, and I've also written some `method` bodies where the statements do turn gray (even for assignments to non-ghost out-parameters, which are definitely not ghost statements).

    Here's one example:

    ![image](https://user-images.githubusercontent.com/11763367/151882127-209826fc-3b40-4f5c-9d94-c55948d6628d.png)

    So, my assessment is now that the feature is rather unreliable. I don't understand when it's recomputing this information or what to expect.

* The other issue I have is selected code doesn't show up over the gray ghost highlighting. Usually, when I select some code, it lights up with a blue background. But when the code is displayed as ghost, then the background remains gray even when I select some code. This feels wrong to me as a user. I would like the blue selection highlighting to take priority over the gray ghost highlighting.

MikaelMayer:
Thanks for reporting. I'm in favor of these changes too.

camrein:
Thanks a lot for reporting.
It's definitely undesirable that the ghost information is not reliable, and even shows wrong information.
I'll try to have a look by myself. If anyone wants to give it a try, see below for things that came to my mind that might be helpful.

The decision whether something is marked as ghost or not is straightforward:
https://github.com/dafny-lang/dafny/blob/614425281b4374a47dbc824822728080ff4e5bd8/Source/DafnyLanguageServer/Language/GhostStateDiagnosticCollector.cs#L59-L62

The first thing I'd suggest to avoid marking lemmas as ghost unnecessarily, is to add a simple filter to the visitor implementation. Then we can avoid traversing down inside lemmas (and functions as well) and won't mark ghost statements inside these.

The thing about user interaction with code selection is also very undesirable. It might be possible to resolve this by changing the color of the ghost highlighting to a transparent one:
https://github.com/dafny-lang/ide-vscode/blob/9ff58873f16ceaa032e8478d6109e84e995c8a62/src/ui/ghostDiagnosticsView.ts#L8-L15

PR #141, opened 2022-03-03
fix: Occasionally missing red squiggles under errors

camrein:
This PR intends to resolve/mitigate #139. However, I couldn't reproduce the issue myself so I can't confirm the fix myself. Moreover, it may improve the mixed experience reported in #126.
Instead of creating a new decoration object with every request, only a single one is created.


MikaelMayer:
Indeed, good catch ! If you don't remove previous decorations, VSCode will just migrate them and never remove them from the code (you'd have to assign them empty ranges for that).

MikaelMayer (approved):

PR #147, opened 2022-03-24
Verification diagnostics highlights

MikaelMayer:
This PR will make the result of this Dafny PR available: 
https://github.com/dafny-lang/dafny/pull/1923

Features implemented:
- Renders the linear verification diagnostics in the left gutter with static and animated icons
  - Verified/Error/Obsolete/Range/Assertion verifier over error/pending
- That same diagnostics are also rendered in the right scrollbar for a preview of the entire file
- Dynamically highlight errors on the line and their related errors
- Option to deactivate verification diagnostics (on by default)
- Option to activate advanced highlighting of verification diagnostics positions (off by default)
- Parse errors will gray out the gutter display after 2 seconds the first time, and immediately after every subsequent resolution or parsing errors
- The first "obsolete" line diagnostics is not shown until 2 seconds after it is first set up, so that there is less flickering for the current method being edited

Needs before merge:
- Remove commented out code


.vscode/settings.json
```diff
@@ -11,5 +11,6 @@
   },
   // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
   "typescript.tsc.autoDetect": "off",
-  "typescript.preferences.quoteStyle": "single"
+  "typescript.preferences.quoteStyle": "single",
+  "dafny-vscode.trace.server": "verbose"
```
camrein, line 15: Do we want this by default in the development experience? To me it makes sense, but might be an accidental commit.

src/language/api/verificationDiagnostics.ts
```diff
@@ -0,0 +1,85 @@
...
+export interface IVerificationDiagnosticsParams {
+  uri: DocumentUri;
+  version?: integer;
+  diagnostics: Diagnostic[];
+  perNodeDiagnostic: NodeDiagnostic[];
+  diagnosticsAreResolutionErrors: boolean;
+  perLineDiagnostic: LineVerificationStatus[];
+}
+
+export interface NodeDiagnostic {
```
camrein, line 13:
```suggestion
export interface INodeDiagnostic {
```

Prepend with `I`.

src/language/api/verificationDiagnostics.ts
```diff
@@ -0,0 +1,85 @@
...
+  displayName: string;
+  identifier: string;
+  position: Position;
+  started: boolean;
+  finished: boolean;
+  startTime: integer;
+  endTime: integer;
+  timeSpent: integer;
```
camrein, line 21: In general, this interface has  quite the number of properties. Is it possible to aggregate some of them to at least avoid primitive obsession? I feel that these three values are candidates. If it even was possible to infer `timeSpent` from the other two properties I'd leave it away to reduce the redundancy.

src/language/api/verificationDiagnostics.ts
```diff
@@ -0,0 +1,85 @@
...
+  resourceCount: integer;
+  range: Range;
+  children: NodeDiagnostic[];
+  statusCurrent: CurrentStatus;
```
camrein, line 25: What's the reason for this name mismatch between this property and its type?

src/language/api/verificationDiagnostics.ts
```diff
@@ -0,0 +1,85 @@
...
+  statusVerification: VerificationStatus;
+  relatedRanges: Range[];
+  immediatelyRelatedRanges?: Range[];
+  dynamicallyRelatedRanges?: Range[];
+}
+
+export enum LineVerificationStatus {
```
camrein, line 32: Do these enumeration numbers come from boogie? I suggest to include it in the type-name or at least in a comment to make this clear.

src/ui/verificationDiagnosticsView.ts
```diff
@@ -0,0 +1,701 @@
...
+  private readonly grayedeDecorations: DecorationSet;
+  private readonly relatedDecorations: TextEditorDecorationType;
+  private readonly relatedDecorationsPartial: TextEditorDecorationType;
+  private readonly relatedDecorationsPartialActive: TextEditorDecorationType;
+  private readonly secondaryRelatedDecorations: TextEditorDecorationType;
+
+  private readonly textEditorWatcher?: Disposable;
+
+  private readonly dataByDocument = new Map<string, LinearVerificationDiagnostics>();
+  private animationCallback: any = 0;
```
camrein, line 53: Don't use `any`.

src/ui/verificationDiagnosticsView.ts
```diff
@@ -0,0 +1,701 @@
+/* eslint-disable max-depth */
```
camrein, line 1: This file has lots of commented code. Is it still work in progress? I'll happily look into it in more detail after it has been cleaned. Even without the commented code file consists of over 500+ lines. There might be opportunities to extract some of the code into separate modules.

camrein (changes requested):
This feature looks fantastic. I've provided you with some intermediate feedback. I'll have to look into it in more detail later.

src/language/api/verificationDiagnostics.ts
MikaelMayer: We are also discussing the name "NodeDiagnostic" here
https://github.com/dafny-lang/dafny/pull/1942#discussion_r838571564
Meanwhile, I'll prefix it with an I

src/language/api/verificationDiagnostics.ts
MikaelMayer: TimeSpent is not completely determined by EndTime and StartTime. The formula in the backend is:
```
public int TimeSpent => (int)(Finished ? ((TimeSpan)(EndTime - StartTime)).TotalMilliseconds : Started ? (DateTime.Now - StartTime).TotalMilliseconds : 0);
```
But otherwise, there is no possibilities to add methods to TypeScript interfaces, is it?
Moreover these fields are sent anyway, and I plan to later use them to display something cool in the back-end (like a gant diagram of the proofs)

src/language/api/verificationDiagnostics.ts
MikaelMayer: I'm adding this comment:
```
// Except for cosmetics, this enumeration is a copy-paste from Dafny's
// Source/DafnyLanguageServer/Workspace/Notifications/VerificationDiagnosticsParams.cs
```

src/ui/verificationDiagnosticsView.ts
MikaelMayer: Yes, I will move this commented out code in another branch.

src/language/api/verificationDiagnostics.ts
camrein: Ah that's not necessary since these modules are supposed to represent the language server's public API. I was wondering if this was a custom enumeration or one that basically represents the states observable within boogie since it's a rather huge enumeration. But I suppose these are all states you're setting manually.

src/language/api/verificationDiagnostics.ts
camrein: Bear in mind that the interfaces we use here roughly describe the format of a JSON object we receive from the backend and all these type additions are not present at runtime. If we would have something like default-methods, it's probably hard for TypeScript to decide when to use the default method and one from the actual object that was assigned to the variable requiring this interface, to not say impossible. In contrast to other languages, TypeScript does not require an instance to actually implement this interface but only have the shape of it.
Unfortunately, TypeScript does not (yet) have extension methods as we have it in C#.

One last note: That the received JSON object reflects the expected shape described in the interface is nowhere (statically) verified, it could be anything.

MikaelMayer: Indeed ! And based on your comment, I did a slight change now: Instead of saying that the interface returns `Range`, which is problematic because it does not contain methods (it's just a JSON object), I say it returns an `IRange`. I was already doing the conversion for every range, but now it's clearer why it is needed.

src/language/api/verificationDiagnostics.ts
```diff
@@ -0,0 +1,83 @@
...
+  line: number;
+  character: number;
+}
+
+export interface IRange {
+  start: IPosition;
+  end: IPosition;
+}
+
+export interface IDiagnostic {
```
keyboardDrummer, line 13: Is this used?

keyboardDrummer:
Nitpick: could you update the PR title/description?

src/constants.ts
```diff
@@ -24,6 +24,8 @@ export namespace ConfigurationConstants {
     export const VerificationCachingPolicy = 'verificationCachingPolicy';
     export const MarkGhostStatements = 'markGhostStatements';
     export const DafnyPlugins = 'dafnyPlugins';
+    export const DisplayGutterStatus = 'displayGutterStatus';
+    export const DisplayVerificationTrace = 'displayVerificationTrace';
```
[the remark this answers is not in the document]
MikaelMayer: Still need to move it to another PR.

keyboardDrummer (approved):

keyboardDrummer:
Looks good, except for the failing build ^^ 

One thing that I'd really like to have is telemetry for this extension, so we could monitor what options users are using. The main difficulty with telemetry is that I'm not sure what service we would use to send the data to.

src/language/api/verificationGutterStatusParams.ts
```diff
@@ -0,0 +1,77 @@
+import { DocumentUri, integer } from 'vscode-languageclient';
+
+export interface IVerificationDiagnosticsParams {
```
keyboardDrummer, line 3: I would include `Gutter` in this name and leave out `Diagnostics` (since it refers to LSP's diagnostics)

src/ui/verificationGutterStatusView.ts
```diff
@@ -0,0 +1,345 @@
...
+    return r;
+  }
+
+  // Given current data and previous data, should we delay the display of the new data?
+  private mustBeDelayed(ranges: Map<LineVerificationStatus, Range[]>, previousRanges: Map<LineVerificationStatus, Range[]>): boolean {
+    return (this.getRanges(ranges, LineVerificationStatus.ResolutionError).length >= 1
+            && this.getRanges(previousRanges, LineVerificationStatus.ResolutionError).length === 0)
+        || (obsoleteLineVerificationStatus.some(status => this.getRanges(ranges, status).length >= 1)
+            && verifyingLineVerificationStatus.every(status => this.getRanges(ranges, status).length === 0)
+            && obsoleteLineVerificationStatus.every(status => this.getRanges(previousRanges, status).length === 0));
```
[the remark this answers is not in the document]
MikaelMayer: Great idea.

src/ui/verificationGutterStatusView.ts
```diff
@@ -0,0 +1,345 @@
...
+      clearInterval(this.animationCallback);
+    }
+    const previousRanges = previousValue === undefined ? VerificationGutterStatusView.emptyLinearVerificationDiagnostics : previousValue.decorations;
+    if(this.mustBeDelayed(ranges, previousRanges)) {
+      // Delay resolution errors so that we don't interrupt the verification workflow if they are corrected fast enough.
+      this.animationCallback = setTimeout(() => {
+        this.dataByDocument.set(documentPath, newData);
+        this.refreshDisplayedVerificationGutterStatuses(window.activeTextEditor);
+      }, DELAY_IF_RESOLUTION_ERROR);
+    } else {
```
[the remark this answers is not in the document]
MikaelMayer: Thanks for the tip, that's interesting ! I did not know about the debounce concept.
This case seems more complicated than debouncing because there are animations too, and timeout is not constant.
I did refactor the code though so that it's more legible. Also, I don't think async/await would increase legibility there, to me the counter-example implementation seems cryptic compared to a simple setTimeout/setInterval I implemented there.

src/ui/verificationGutterStatusView.ts
```diff
@@ -0,0 +1,376 @@
...
+            : path.startsWith('error')
+              ? ScrollColor.Error
+              : path.startsWith('verified')
+                ? ScrollColor.Verified
+                : ScrollColor.Unknown
+    });
+  }
+
+  /// Helper to create decoration types based on the mode and number of images
+  public makeIconAux(grayMode: boolean, context: ExtensionContext):
```
camrein, line 141: I think this function could either be declared either static or outside of the class.

camrein (approved):
looks great 👍

PR #149, opened 2022-03-29
Fix: Avoid out-of-order messages

MikaelMayer:
I witnessed that sometimes verification is not canceled in a timely manner, leading to obsolete notifications to be received by the client. This PR ensures that obsolete notifications are not sent if notifications for a newer document is available.

### DIFF
--- a/src/ui/compilationStatusView.ts
+++ b/src/ui/compilationStatusView.ts
@@ -27,10 +27,15 @@ function toStatusMessage(status: CompilationStatus, message?: string | null): st
   }
 }
 
+interface DocumentStatusMessage {
+  status: string;
+  version: number | undefined;
+}
+
 export default class CompilationStatusView {
   // We store the message string for easier backwards compatibility with the
   // legacy status messages.
-  private readonly documentStatusMessages = new Map<string, string>();
+  private readonly documentStatusMessages = new Map<string, DocumentStatusMessage>();
 
   private constructor(private readonly statusBarItem: StatusBarItem) {}
 
@@ -55,18 +60,32 @@ export default class CompilationStatusView {
     this.updateActiveDocumentStatus();
   }
 
+  private areParamsOutdated(params: ICompilationStatusParams): boolean {
+    const previous = this.documentStatusMessages.get(getVsDocumentPath(params));
+    return previous?.version !== undefined && params.version !== undefined
+       && previous.version > params.version;
+  }
+
   private compilationStatusChanged(params: ICompilationStatusParams): void {
+    if(this.areParamsOutdated(params)) {
+      return;
+    }
     this.documentStatusMessages.set(
       getVsDocumentPath(params),
-      toStatusMessage(params.status, params.message)
+      { status: toStatusMessage(params.status, params.message),
+        version: params.version
+      }
     );
     this.updateActiveDocumentStatus();
   }
 
+  // Legacy methods
   private verificationStarted(params: IVerificationStartedParams): void {
     this.documentStatusMessages.set(
       getVsDocumentPath(params),
-      Messages.CompilationStatus.Verifying
+      { status: Messages.CompilationStatus.Verifying,
+        version: undefined
+      }
     );
     this.updateActiveDocumentStatus();
   }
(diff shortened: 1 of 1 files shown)


src/ui/compilationStatusView.ts
```diff
@@ -27,10 +27,15 @@ function toStatusMessage(status: CompilationStatus, message?: string | null): st
   }
 }
 
+interface DocumentStatusMessage {
+  status: string;
+  version: number | undefined;
```
cpitclaudel, line 32:
```suggestion
  version?: number;
```

src/ui/compilationStatusView.ts
```diff
@@ -55,26 +60,42 @@ export default class CompilationStatusView {
...
     this.updateActiveDocumentStatus();
   }
 
   private verificationCompleted(params: IVerificationCompletedParams): void {
     this.documentStatusMessages.set(
       getVsDocumentPath(params),
-      params.verified ? Messages.CompilationStatus.Verified : Messages.CompilationStatus.NotVerified
+      { status: params.verified ? Messages.CompilationStatus.Verified : Messages.CompilationStatus.NotVerified,
+        version: undefined
+      }
```
cpitclaudel, line 98:
```suggestion
      { status: params.verified ? Messages.CompilationStatus.Verified : Messages.CompilationStatus.NotVerified }
```

MikaelMayer:
I just realized it could also be why sometimes we see a message "Verifying...." even when something is done verified, because of out of order notifications.

MikaelMayer:
I'll wait for @camrein to have a look at it as well.

src/ui/compilationStatusView.ts
```diff
@@ -27,10 +27,15 @@ function toStatusMessage(status: CompilationStatus, message?: string | null): st
   }
 }
 
+interface DocumentStatusMessage {
```
camrein, line 30:
```suggestion
interface IDocumentStatusMessage {
```
So far we've got the convention to prefix interfaces with `I`.

src/ui/compilationStatusView.ts
```diff
@@ -55,26 +60,37 @@ export default class CompilationStatusView {
...
+
   private compilationStatusChanged(params: ICompilationStatusParams): void {
+    if(this.areParamsOutdated(params)) {
+      return;
+    }
     this.documentStatusMessages.set(
       getVsDocumentPath(params),
-      toStatusMessage(params.status, params.message)
+      { status: toStatusMessage(params.status, params.message),
+        version: params.version }
```
camrein, line 76:
```suggestion
      {
         status: toStatusMessage(params.status, params.message),
         version: params.version
      }
```

src/ui/compilationStatusView.ts
```diff
@@ -55,26 +60,37 @@ export default class CompilationStatusView {
     this.updateActiveDocumentStatus();
   }
 
+  private areParamsOutdated(params: ICompilationStatusParams): boolean {
+    const previous = this.documentStatusMessages.get(getVsDocumentPath(params));
+    return previous?.version !== undefined && params.version !== undefined
```
camrein, line 65:
```suggestion
    return previous?.version != null && params.version != null
```

nit: Although it's no longer problematic to compare against `undefined`, the code-base uses it interchangeably with `null` so far. See for example:

https://github.com/dafny-lang/ide-vscode/blob/d4ce528808aaaeaaf839ff11b8965f005887fef8/src/ui/dafnyVersionView.ts#L28

and

https://github.com/dafny-lang/ide-vscode/blob/2eec391d39ca71057c3ec92130d987ec4b5e8f72/src/tools/debounce.ts#L16-L19

But I'm open to refactorings if there is some need or preference to be more strict here.

camrein (changes requested):
Nice catch! And thanks @cpitclaudel for your great input as well!

src/ui/compilationStatusView.ts
MikaelMayer: Oh I did not see it would be accepted by the linter. Nice. I used your expression.

src/ui/compilationStatusView.ts
```diff
@@ -55,26 +60,38 @@ export default class CompilationStatusView {
...
   private compilationStatusChanged(params: ICompilationStatusParams): void {
+    if(this.areParamsOutdated(params)) {
+      return;
+    }
     this.documentStatusMessages.set(
       getVsDocumentPath(params),
-      toStatusMessage(params.status, params.message)
+      { status: toStatusMessage(params.status, params.message),
+        version: params.version
+      }
```
camrein, line 77:
```suggestion
      {
        status: toStatusMessage(params.status, params.message),
        version: params.version
      }
```
Have `status` on a new-line too. 👍

camrein (changes requested):
Great addition. Just refine the small formatting issue and feel free to merge afterwards.

camrein (approved):

Issue #146, opened 2022-03-18
Problems running Dafny on Apple M1 devices


MikaelMayer:
On helping someone install VSCode, dotnet 6, and then Dafny, they get the following error:

```
Failed to load /Users/<user>/.vscode/extensions/dafny-lang.ide-vscode-2.3.0/out/resources/3.5.0/dafny/libhostpolicy.dylib, error: dlopen(/Users/<user>/.vscode/extensions/dafny-lang.ide-vscode-2.3.0/out/resources/3.5.0/dafny/libhostpolicy.dylib, 1): no suitable image found.  Did find:
    /Users/<user>/.vscode/extensions/dafny-lang.ide-vscode-2.3.0/out/resources/3.5.0/dafny/libhostpolicy.dylib: mach-o, but wrong architecture
    /Users/<user>/.vscode/extensions/dafny-lang.ide-vscode-2.3.0/out/resources/3.5.0/dafny/libhostpolicy.dylib: mach-o, but wrong architecture
An error occurred while loading required library libhostpolicy.dylib from [/Users/<user>/.vscode/extensions/dafny-lang.ide-vscode-2.3.0/out/resources/3.5.0/dafny/]
[Info  - 9:57:07 AM] Connection to server got closed. Server will restart.
(... 5 times)
[Error - 9:57:07 AM] Connection to server got closed. Server will not be restarted.
```

Not sure where this come from

camrein:
Sounds to me like the binary that is downloaded the mac (see the `dylib` extension) is not compatible, see `mach-o, but wrong architecture`.
Dafny ` dafny-?.?.?-x64-osx-10.14.2.zip` is an `x86_64` distro. Does the mac happen to be an ARM / Apple M1 device? Then I suppose we need to distribute suitable OSX images, i.e., execute `dotnet publish` with the respective architecture:

See:
https://github.com/dafny-lang/dafny/blob/b133a187c80e26c25af32bdb6488e3a40e631f41/Scripts/package.py#L137-L143

And:
https://docs.microsoft.com/en-us/dotnet/core/rid-catalog#macos-rids

We'll need to pass the architecture to the `publish` command, for example:
```sh
dotnet publish -r osx-arm64
```

camrein:
> I don't recall whether I did any particular setup, but I might have installed this x64 architecture emulator: https://support.apple.com/en-us/HT211861

[...]

PR #155, opened 2022-05-11
fix: Solves the problem of OSX M1 architectures.

### LINKED ISSUE #148: Is dafny-vscode extension working for M1 Mac?
Is it possible to install dafny-vscode in M1 mac?

I installed .NET 6.0, and then the extension of dafny, but I received the following error in the output of the terminal.

`Failed to load /Users/<User>/.vscode/extensions/dafny-lang.ide-vscode-2.3.0/out/resources/3.5.0/dafny/libhostpolicy.dylib, error: dlopen(/Users/<User>/.vscode/extensions/dafny-lang.ide-vscode-2.3.0/out/resources/3.5.0/dafny/libhostpolicy.dylib, 0x0001): tried: '/Users/<User>/.vscode/extensions/dafny-lang.ide-vscode-2.3.0/out/resources/3.5.0/dafny/libhostpolicy.dylib' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e'))
An error occurred while loading required library libhostpolicy.dylib from [/Users/<User>/.vscode/extensions/dafny-lang.ide-vscode-2.3.0/out/resources/3.5.0/dafny/]
[Info  - 8:42:59 PM] Connection to server got closed. Server will restart.`

MikaelMayer:
Fixes #148 and #146 in a custom way.

This PR adds custom "manual" installation for an architecture that is not supported during the release process.
I tried to follow the steps in https://github.com/dafny-lang/dafny/wiki/INSTALL for mac, but I omitted the following steps:
- Installing `.NET 6.0`. Is this necessary, or will users have .NET 6.0 installed anyway because they'd have vs code? I was thinking the latter.
- Installing `pre-commit`. I don't think it's useful for them.
- Installing `wget`, as it's likely part of their machines.

This extension was tested on a mac M1 machine and works the same as installing Dafny manually and pointing to it.
To test this PR, run `npx vsce package` on a folder with this PR checked out, and use VSCode extension manager to install this extension.

However, some problems remain that appear on these M1 machines that don't appear on windows, but it would be another problem to solve I believe. To reproduce this, a user would open their VSCode on [this file:](https://github.com/dafny-lang/dafny/blob/master/Test/comp/Arrays.dfy), position their caret at the end of line 381, type ENTER, write character by character `assert false;` and remove it character by character. VSCode gets stucks eventually.

### DIFF
--- a/src/language/dafnyInstallation.ts
+++ b/src/language/dafnyInstallation.ts
@@ -114,13 +114,43 @@ export class DafnyInstaller {
       await this.extractArchive(archive);
       await workspace.fs.delete(archive, { useTrash: false });
       this.writeStatus('Dafny installation completed');
-      return true;
+      if(os.type() === 'Darwin' && os.arch() !== 'x64') {
+        // Need to build from source and move all files from Binary/ to the current foldre
+        this.writeStatus('Found a non-supported architecture OSX:' + os.arch() + '. Going to install from source and replace the automated installation.\n');
+        return await this.installFromSource();
+      } else {
+        return true;
+      }
     } catch(error: unknown) {
       this.writeStatus('Dafny installation failed:');
       this.writeStatus(`> ${error}`);
       console.error('dafny installation failed', error);
       return false;
     }
+  }
+
+  private async installFromSource() {
+    const { execSync } = require('child_process');
+    const execLog = (command: string) => {
+      this.writeStatus(`Executing: ${command}\n`);
+      execSync(command);
+    };
+    const installationPath = this.getCustomInstallationPath(os.arch());
+    await mkdirAsync(installationPath.fsPath, { recursive: true });
+    this.writeStatus(`Installing Dafny from source in ${installationPath.fsPath}.\n`);
+    const process = require('process');
+    process.chdir(installationPath.fsPath);
+    execLog('git clone --recurse-submodules https://github.com/dafny-lang/dafny.git');
+    process.chdir(Utils.joinPath(installationPath, 'dafny').fsPath);
+    execLog('make exe');
+    const Binaries = Utils.joinPath(installationPath, 'dafny', 'Binaries').fsPath;
+    process.chdir(Binaries);
+    execLog('wget https://github.com/Z3Prover/z3/releases/download/Z3-4.8.5/z3-4.8.5-x64-osx-10.14.2.zip');
+    execLog('unzip z3-4.8.5-x64-osx-10.14.2.zip');
+    execLog('mv z3-4.8.5-x64-osx-10.14.2 z3');
+    process.chdir(this.getInstallationPath().fsPath);
+    execLog(`cp -R ${Binaries}/* ./dafny/`);
+    return true;
   }
 
   public isCustomInstallation(): boolean {
@@ -193,6 +223,13 @@ export class DafnyInstaller {
     return Utils.joinPath(
       this.context.extensionUri,
       ...LanguageServerConstants.GetResourceFolder(getConfiguredVersion())
+    );
+  }
+
+  private getCustomInstallationPath(typeArch: string): Uri {
+    return Utils.joinPath(
+      this.getInstallationPath(),
+      ...[ 'custom', typeArch ]
     );
   }
 


cpitclaudel:
> will users have .NET 6.0 installed anyway because they'd have vs code?

VSCode is written in TypeScript, right?  I don't think it depends on dotnet.

cpitclaudel:
Very nice job!  Congrats @prvshah51 and @MikaelMayer .

But I'm confused: if it works to build locally, why can't we do it in CI and release a package for that architecture as well?

MikaelMayer:
> But I'm confused: if it works to build locally, why can't we do it in CI and release a package for that architecture as well?

It should be possible in theory, but both @prvshah51 and I were confused about the release process, I'm myself working on a Windows machine and we could not figure it out the first hour we worked together. So we did what we could best, which was to build locally from vs code. I would be very happy if someone could update the release process of Dafny, I would know how to update the VSCode extension to take it into account.

src/language/dafnyInstallation.ts
```diff
@@ -123,6 +129,32 @@ export class DafnyInstaller {
     }
   }
 
+  private async installFromSource() {
+    const { execSync } = require('child_process');
```
camrein, line 133: Don't use `require` within ES Modules. As well as avoid using blocking/sync functions whenever possible. You may want to have a look here:
https://github.com/dafny-lang/ide-vscode/blob/cf35f2c4539bb75d102f7219ef185d4bdd5b0ff7/src/dotnet.ts#L15

It should help you solve both things, making an asynchronous call and using the `import` statement.

src/language/dafnyInstallation.ts
```diff
@@ -114,7 +114,13 @@ export class DafnyInstaller {
       await this.extractArchive(archive);
       await workspace.fs.delete(archive, { useTrash: false });
       this.writeStatus('Dafny installation completed');
-      return true;
+      if(os.type() === 'Darwin' && os.arch() !== 'x64') {
+        // Need to build from source and move all files from Binary/ to the current foldre
+        this.writeStatus('Found a non-supported architecture OSX:' + os.arch() + '. Going to install from source and replace the automated installation.\n');
```
camrein, line 119: Prefer the use of format strings. Line-breaks are automatically appended (easier for OS interoperability):
```suggestion
        this.writeStatus(`Found a non-supported architecture OSX: ${os.arch()}. Going to install from source and replace the automated installation.`);
```

src/language/dafnyInstallation.ts
```diff
@@ -123,6 +129,32 @@ export class DafnyInstaller {
     }
   }
 
+  private async installFromSource() {
+    const { execSync } = require('child_process');
+    const execLog = (command: string) => {
+      this.writeStatus(`Executing: ${command}\n`);
+      execSync(command);
+    };
```
camrein, line 137: When applying my above suggestion, you no longer need to inline this code. Again, the line-break should be redundant.

src/language/dafnyInstallation.ts
```diff
@@ -123,6 +129,32 @@ export class DafnyInstaller {
...
+    const installationPath = this.getCustomInstallationPath(os.arch());
+    await mkdirAsync(installationPath.fsPath, { recursive: true });
+    this.writeStatus(`Installing Dafny from source in ${installationPath.fsPath}.\n`);
+    const process = require('process');
```
camrein, line 141: Prefer the use of the `import` statement.

src/language/dafnyInstallation.ts
```diff
@@ -196,6 +228,13 @@ export class DafnyInstaller {
     );
   }
 
+  private getCustomInstallationPath(typeArch: string): Uri {
+    return Utils.joinPath(
+      this.getInstallationPath(),
+      ...[ 'custom', typeArch ]
+    );
```
camrein, line 235: This should be sufficient:
```suggestion
    return Utils.joinPath(this.getInstallationPath(), 'custom', typeArch);
```

src/language/dafnyInstallation.ts
```diff
@@ -123,6 +129,32 @@ export class DafnyInstaller {
...
+    process.chdir(Utils.joinPath(installationPath, 'dafny').fsPath);
+    execLog('make exe');
+    const Binaries = Utils.joinPath(installationPath, 'dafny', 'Binaries').fsPath;
+    process.chdir(Binaries);
+    execLog('brew install wget');
+    execLog('wget https://github.com/Z3Prover/z3/releases/download/Z3-4.8.5/z3-4.8.5-x64-osx-10.14.2.zip');
+    execLog('unzip z3-4.8.5-x64-osx-10.14.2.zip');
+    execLog('mv z3-4.8.5-x64-osx-10.14.2 z3');
+    process.chdir(this.getInstallationPath().fsPath);
+    execLog(`cp -R ${Binaries}/* ./dafny/`);
```
camrein, line 154: You may want to check the return codes of the commands to not proceed when any step fails. On top of that you could provide some meaningful help message.

src/language/dafnyInstallation.ts
```diff
@@ -123,6 +129,32 @@ export class DafnyInstaller {
...
+    const installationPath = this.getCustomInstallationPath(os.arch());
+    await mkdirAsync(installationPath.fsPath, { recursive: true });
+    this.writeStatus(`Installing Dafny from source in ${installationPath.fsPath}.\n`);
+    const process = require('process');
+    process.chdir(installationPath.fsPath);
+    execLog('brew install dotnet-sdk');
+    execLog('git clone --recurse-submodules https://github.com/dafny-lang/dafny.git');
+    process.chdir(Utils.joinPath(installationPath, 'dafny').fsPath);
+    execLog('make exe');
+    const Binaries = Utils.joinPath(installationPath, 'dafny', 'Binaries').fsPath;
```
camrein, line 147: Start local variables with a small letter.

src/language/dafnyInstallation.ts
```diff
@@ -123,6 +129,32 @@ export class DafnyInstaller {
...
+    const { execSync } = require('child_process');
+    const execLog = (command: string) => {
+      this.writeStatus(`Executing: ${command}\n`);
+      execSync(command);
+    };
+    const installationPath = this.getCustomInstallationPath(os.arch());
+    await mkdirAsync(installationPath.fsPath, { recursive: true });
+    this.writeStatus(`Installing Dafny from source in ${installationPath.fsPath}.\n`);
+    const process = require('process');
+    process.chdir(installationPath.fsPath);
```
camrein, line 142: Isn't it possible to avoid changing the working directory, i.e., adapt the commands accordingly? Then you no longer need the `process` module as well as do not mess with VSCodes working directory. If not, you should at least switch back to the original working directory. Otherwise, this could lead to undesired bugs that would probably hard for us to reproduce.

camrein (changes requested):
This looks great! I've added a bunch of notes that could be refined.

src/language/dafnyInstallation.ts
MikaelMayer: What do you mean by "need to inline this code"?

src/language/dafnyInstallation.ts
MikaelMayer: I don't think this is necessary, as `exec` will throw an exception if the return code is not zero.
https://stackoverflow.com/questions/32874316/node-js-accessing-the-exit-code-and-stderr-of-a-system-command

BhuvanVankadari:
I'm sorry if this is the wrong place to post this, but I don't seem to understand the method to get the extension working for M1 Mac, could you please elaborate on that?

prvshah51:
You just have to install the extension from vscode. There is no need to do anything else.

BhuvanVankadari:
I did install the extension from VSCode but I get the error `An error occurred while loading required library libhostpolicy.dylib`

prvshah51:
Can you open a new issue in the repo and paste all the logs there? It will be helpful to keep issues you are having at one place to find the solution, Thanks.

PR #159, opened 2022-05-20
Better error if dotnet is not installed or not configured properly

### LINKED ISSUE #112: checkDotnetInstallation doesn't distinguish between an incorrect dotnet path and invalid version
Dear Dafny team,

While splunking around during work on #111 I noticed my extension settings were slightly misconfigured: I had misread the description of `dafny.dotnetExecutablePath` to say "the absolute path _to_ dotnet", rather than "the absolute path _of_ dotnet", and I supplied the directory that contains the executable rather than the executable itself.  A simple mistake, easily fixed.  However, in that situation, the `NoCompatibleInstallation` message is still presented to the user, which, at least to my reading, suggests that `dotnet` _was_ found but determined to be an incorrect version, whereas in reality the plugin couldn't find it at all. 

To my mind, the issue is that `hasSupportedDotnetVersion()` eats any exception raised by `execFileAsync()` and simply returns `false` if _anything_ goes wrong, so the UI can't distinguish between an invalid core runtime string being parsed out of running `dotnet` and a failure to run it at all.  I might suggest enriching the return type of `hasSupportedDotnetVersion` for the UI to be able to distinguish this.  Ideally, from the user's perspective, they would be presented different messages in the following cases:

* `dotnet` does not have a valid runtime
* `dotnet` is found but not executable
* `dotnet` is not a file / is missing
* `dotnetExecutable` does not encode a valid path

I imagine the last three can be extracted from the caught exception.  But, I think really, distinguishing between "you haven't given me a valid `dotnet`" versus "`dotnet` is not the right version" would be more than enough.

I have a bunch of spare cycles owing to the holidays, so, if you agree that this is a reasonable enhancement I'm happy to try to take a stab at it.  Thanks!

MikaelMayer:
Fixes #112 

If we specify a directory
image

If we specify an executable that is not a compatible dotnet file
image

If we specify a _file_ that is not executable (sic!), at least we don't get the "no compatible dotnet installed", we get an explicit error about what happened
image

Finally, I changed the default error message from `There is no compatible dotnet runtime installed` to `No compatible dotnet runtime found`, and this message is shown only if the user did not specify an explicit dotnet path.

### DIFF
--- a/src/dotnet.ts
+++ b/src/dotnet.ts
@@ -16,7 +16,7 @@ const lstatAsync = promisify(fs.lstat);
 export async function checkSupportedDotnetVersion(): Promise<string | undefined> {
   const { path: dotnetExecutable, manual } = await getDotnetExecutablePath();
   try {
-    const stats = await lstatAsync(dotnetExecutable) as fs.Stats;
+    const stats = await lstatAsync(dotnetExecutable);
     if(!stats.isFile()) {
       return dotnetExecutable + Messages.Dotnet.IsNotAnExecutableFile;
     }
--- a/src/dotnet.ts
+++ b/src/dotnet.ts
@@ -1,34 +1,42 @@
 import { execFile } from 'child_process';
 import { promisify } from 'util';
+import * as fs from 'fs';
 
 import * as which from 'which';
 
 import { ConfigurationConstants, DotnetConstants } from './constants';
 import Configuration from './configuration';
+import { Messages } from './ui/messages';
 
 const ListRuntimesArg = '--list-runtimes';
 const execFileAsync = promisify(execFile);
 
-export async function hasSupportedDotnetVersion(): Promise<boolean> {
-  const dotnetExecutable = await getDotnetExecutablePath();
+// Returns an error message
+export async function hasSupportedDotnetVersion(): Promise<string> {
+  const { path: dotnetExecutable, manual } = await getDotnetExecutablePath();
   try {
+    if(!fs.lstatSync(dotnetExecutable).isFile()) {
+      return dotnetExecutable + Messages.Dotnet.IsNotAnExecutableFile;
+    }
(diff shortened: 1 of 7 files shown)


src/dotnet.ts
```diff
@@ -1,34 +1,42 @@
...
 import Configuration from './configuration';
+import { Messages } from './ui/messages';
 
 const ListRuntimesArg = '--list-runtimes';
 const execFileAsync = promisify(execFile);
 
-export async function hasSupportedDotnetVersion(): Promise<boolean> {
-  const dotnetExecutable = await getDotnetExecutablePath();
+// Returns an error message
+export async function hasSupportedDotnetVersion(): Promise<string> {
```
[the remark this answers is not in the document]
MikaelMayer: I  renamed it `checkSupportedDotnetVersion` and it now returns `Promise<string | undefined>`

src/dotnet.ts
```diff
@@ -1,34 +1,44 @@
...
 const execFileAsync = promisify(execFile);
+const lstatAsync = promisify(fs.lstat);
 
-export async function hasSupportedDotnetVersion(): Promise<boolean> {
-  const dotnetExecutable = await getDotnetExecutablePath();
+// Returns an error message or undefined.
+export async function checkSupportedDotnetVersion(): Promise<string | undefined> {
+  const { path: dotnetExecutable, manual } = await getDotnetExecutablePath();
   try {
+    const stats = await lstatAsync(dotnetExecutable) as fs.Stats;
```
camrein, line 19: nit: Instead of using `as`, you may want to parameterize `promisify`. Maybe something like this (untested):
```cs
const lstatAsync = promisify<string, fs.Stats>(fs.lstat);
```

camrein (approved):
great!

src/dotnet.ts
MikaelMayer: Oh actually the cast is not longer needed, I had added it because I was promisifying the wrong version fs.lstatAsync

PR #163, opened 2022-05-23
If language server path "" should be the same as null.

MikaelMayer:
If the user empties the field "language server path", it should be treated the same as if it was null. It's totally unobvious to have to "reset" this parameter.

Before this PR, a setting like this:
image
was setting the language server path to "" and since it was non-null, it would not let the extension download Dafny, for example.

### DIFF
--- a/src/language/dafnyInstallation.ts
+++ b/src/language/dafnyInstallation.ts
@@ -174,7 +174,8 @@ export class DafnyInstaller {
   }
 
   public isCustomInstallation(): boolean {
-    return getConfiguredLanguageServerRuntimePath() != null;
+    const configuredLanguageServerRuntimePath = getConfiguredLanguageServerRuntimePath();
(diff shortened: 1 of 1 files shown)


prvshah51 (approved):

src/language/dafnyInstallation.ts
```diff
@@ -174,7 +174,8 @@ export class DafnyInstaller {
   }
 
   public isCustomInstallation(): boolean {
-    return getConfiguredLanguageServerRuntimePath() != null;
+    const configuredLanguageServerRuntimePath = getConfiguredLanguageServerRuntimePath();
+    return configuredLanguageServerRuntimePath != null && configuredLanguageServerRuntimePath !== '';
```
keyboardDrummer, line 178: I think `!!configuredLanguageServerRuntimePath` would also work.

PR #164, opened 2022-05-24
Feat: Underline related errors in postconditions

MikaelMayer:
Previously, a failing postcondition like `x >= 3` was not underlined in red because it was not a "diagnostic"
image

With this PR, the failing postcondition `x >= 3` is also underlined in orange:
image
The weird indentation is here to visually compare the two underlines, as well as the underline of a warning in yellow just below.

Light mode version:
image

Implementation notes:
* I added a capability in the `DafnyLanguageServer` to hook on the published diagnostics
* I created a class `RelatedErrorView` that uses this hook, looks into diagnostics that start with "A postcondition might not hold", and stores the ranges of the first related informations in a map from document to ranges.
* This class then displays these ranges using the `textDecoration: "underline wavy red 1px"` decoration.
* There is no delay between the display of diagnostics and related errors implemented in this PR.

UX notes:
* I picked up manually the color of regular errors, the color of warnings, and took an average color by color
* I did not find how to make the squiggle look exactly the same, but I'm happy with the result
* With https://github.com/dafny-lang/dafny/pull/1946 which is now merged, hovering this related range also provides a meaningful error message and statistics about the failing postcondition, so users won't be disappointed by hovering them.
* With https://github.com/dafny-lang/ide-vscode/pull/147 which is now merged, the scroll bar on the left will already display a red mark for the line containing the failing postcondition. It will be very important to merge this PR after #147 because otherwise the gutter will display an error in the line but the error won't be visible.


keyboardDrummer:
What's the reason we want to show red squiggles under the return statement? I was thinking we might let the failing ensures clause be the main locations of such verification diagnostics.

For a problem like the following, the return statement doesn't seem relevant:

```dafny
method Foo(x: bool) returns (r: int) ensures x > 1 {
  if (x) {
    r := 2;
  } else {
    r := 0;
  }
  var bar := zap;
}
```

Wouldn't it be better to show the entire error trace in related locations?

cpitclaudel:
> For a problem like the following, 

I don't understand this example; did you mean the following?

```
method Foo(x: bool) returns (r: int) ensures r > 1 {
  if (x) {
    r := 2;
  } else {
    r := 0;
  }
  var bar := "zap";
}
```

I think that's arguably a bug, since the following is equivalent but provides much better diagnostics:

```
method Foo(x: bool) returns (r: int) ensures r > 1 {
  if (x) {
    return 2;
  } else {
    return 0;
  }
  var bar := "zap";
}
```

cpitclaudel:
Btw, this is one case in which the trace is useful (albeit hard to read) on the command line, right?

```
Execution trace:
    (0,0): anon0
    (0,0): anon4_Else
    (0,0): anon3
```

keyboardDrummer:
> > For a problem like the following,
> 
> I don't understand this example; did you mean the following?
> 
> ```
> method Foo(x: bool) returns (r: int) ensures r > 1 {
>   if (x) {
>     r := 2;
>   } else {
>     r := 0;
>   }
>   var bar := "zap";
> }
> ```

yes :) although the `var bar := "zap"` was meant to indicate some operation that should not be skipped in either of the if/then/else cases, so your equivalent program isn't really equivalent because it returns too early.

keyboardDrummer (changes requested):
I would like to discuss what we want the UI to be before merging

MikaelMayer:
> Is the reason for this PR because VSCode doesn't allow configuring showing red squiggles under related information? Have we tried cutting an issue to their repository?

No it's not just VSCode, it's a language server protocol problem: Diagnostics related informations only contain a message and a location, but no indication of whether they should be marked as errors. And it makes sense: You don't want to mark related locations as errors if they are on other functions for example. Errors visualized in the IDE should be only those in context.

> What's the reason we want to show red squiggles under the return statement? I was thinking we might let the failing ensures clause be the main locations of such verification diagnostics.

I talked with the rest of the team today, and the consensus is that it does make sense to leave the main error here, because the ensures clause is asserted there even though it's not visible.
Perhaps it could make sense to put the error on the closing brace rather than the opening brace if applicable, but that would be another discussion.

> For a problem like the following, the return statement doesn't seem relevant:
> 
> ```
> method Foo(x: bool) returns (r: int) ensures x > 1 {
>   if (x) {
>     r := 2;
>   } else {
>     r := 0;
>   }
>   var bar := zap;
> }
> ```
> Wouldn't it be better to show the entire error trace in related locations of the diagnostic?

That's an interesting suggestion. We would first need to prune the error trace because, as you can see, it's [can be pretty messy](https://github.com/dafny-lang/dafny/pull/1977). and in your case, I think it would make sense only if the main error is on the closing brace than the opening brace.
I suggest we keep this idea out of this scope, but that we keep this idea because I think it's promising. After all, a failed assertion's related ranges don't need to be failing sub-expressions, they can indeed display what is the path that was taken if it's not obvious.

keyboardDrummer:
> > Is the reason for this PR because VSCode doesn't allow configuring showing red squiggles under related information? Have we tried cutting an issue to their repository?
> 
> No it's not just VSCode, it's a language server protocol problem: Diagnostics related informations only contain a message and a location, but no indication of whether they should be marked as errors. And it makes sense: You don't want to mark related locations as errors if they are on other functions for example. Errors visualized in the IDE should be only those in context.

To me, how the IDE visualises the different information contained in a diagnostic is up to the IDE. What is the fundamental distinction between an error and related location? I thought error meant "look at this first" and related location meant "look at this second".

> > What's the reason we want to show red squiggles under the return statement? I was thinking we might let the failing ensures clause be the main locations of such verification diagnostics.
> 
> I talked with the rest of the team today, and the consensus is that it does make sense to leave the main error here, because the ensures clause is asserted there even though it's not visible. 

Isn't saying that the ensures clause is asserted before the return statement just one way of interpreting it? Alternatively, you could interpret the return statement as jumping to the ensures clause before jumping to the calling function. If you do that, then you can show the main error on the ensures clause instead of on the "invisible assertion" on the return statement, which I think helps the user because then they don't have to find the ensures clause.

I think a good reason to show the error on the return statement is because it can also provide information about the error trace, but I don't feel that reason is good enough.

PR #167, opened 2022-05-31
Fix: Ghost background in light theme.

### LINKED ISSUE #166: ghost statament color in Dafny on Visual Code
I filed this issue in stackoverflow:

Could I configure the color (grey, by default) of Ghost Statements in Dafny. I mean, the current gray color is too dark, it is difficult to see the text in the classroom screen. I think a lighter gray would be better.

I received the following answers:

-I think it makes more sense to file this as an issue on Github [github.com/dafny-lang/dafny/issues]
[James Wilcox](https://stackoverflow.com/users/438267/james-wilcox)

-I think [github.com/dafny-lang/ide-vscode/issues](https://github.com/dafny-lang/ide-vscode/issues) is right place to file this issue – [Divyanshu Ranjan](https://stackoverflow.com/users/14413069/divyanshu-ranjan)

-There are two options to fix this.
1.- Don't use markGhostStatements. By default dafny.markGhostStatements is set to be true. Change extension setting to false.
2.- Use markGhostStatements, but change background color. There is no configuration to set it. Probably file an issue at https://github.com/dafny-lang/ide-vscode to provide that configuration. You can locally build ide-vscode by setting this [option](https://github.com/dafny-lang/ide-vscode/blob/be6ee7b9dc626401c414d24c32924f6735cf4cf2/src/ui/ghostDiagnosticsView.ts#L13). 
See images below for change in background color by setting GhostDecoration.

image
image

Thank you.

MikaelMayer:
Fixed #166  by making the ghost background color in light theme lighter.

### DIFF
--- a/src/ui/ghostDiagnosticsView.ts
+++ b/src/ui/ghostDiagnosticsView.ts
@@ -10,7 +10,7 @@ const GhostDecoration: DecorationRenderOptions = {
     backgroundColor: '#64646480'
   },
   light: {
-    backgroundColor: '#79797980'
+    backgroundColor: '#D3D3D380'
   }
 };
 


camrein (approved):
nice!

PR #169, opened 2022-06-01
Better error message if java is missing.

MikaelMayer:
Previously, "Make exe" on Mac M1 architectures would be failing silently if java was not present (for the DafnyRuntime dependency)
Now, it raises an explicit error by testing the java version.

### DIFF
--- a/src/dotnet.ts
+++ b/src/dotnet.ts
@@ -17,7 +17,7 @@ export async function checkSupportedDotnetVersion(): Promise<string | undefined>
   const { path: dotnetExecutable, manual } = await getDotnetExecutablePath();
   try {
     const stats = await lstatAsync(dotnetExecutable);
-    if(!stats.isFile()) {
+    if(!stats.isFile() && !stats.isSymbolicLink()) {
       return dotnetExecutable + Messages.Dotnet.IsNotAnExecutableFile;
     }
     const { stdout } = await execFileAsync(dotnetExecutable, [ ListRuntimesArg ]);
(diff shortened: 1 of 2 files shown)


src/dotnet.ts
```diff
@@ -17,7 +17,7 @@ export async function checkSupportedDotnetVersion(): Promise<string | undefined>
   const { path: dotnetExecutable, manual } = await getDotnetExecutablePath();
   try {
     const stats = await lstatAsync(dotnetExecutable);
-    if(!stats.isFile()) {
+    if(!stats.isFile() && !stats.isSymbolicLink()) {
```
MikaelMayer, line 20: Otherwise this was not working on some machines

src/language/dafnyInstallation.ts
```diff
@@ -159,6 +159,19 @@ export class DafnyInstaller {
       this.writeStatus('If you got `brew: command not found`, but brew is installed on your system, please add all brew commands to your ~/.zprofile, e.g. https://apple.stackexchange.com/a/430904 and reinstall Dafny.');
       return false;
     }
+    try {
+      const result = (await this.execLog('javac -version')).stdout;
+      if(!(/javac \d+\.\d+/.exec(result))) {
```
keyboardDrummer, line 164: What do we need java for?

camrein: This is related to the issue I've opened last year: https://github.com/dafny-lang/dafny/issues/1530
At this time, the Dafny runtime requires java to compile, even though the C# distribution doesn't require it. I'd rather fix this issue than introducing another workaround.

camrein: afaik the previous change was to enable M1 support by compiling Dafny from source. Therefore, the `javac` command is required because the `DafnyRuntime` project still relies on it.

PR #170, opened 2022-06-02
Enable configuring a custom language server through an environment variable

keyboardDrummer:
### Changes
- Enable starting the extension with a custom language server using an environment variable, which is useful for development scenarios. You can imagine having a make target in the Dafny repository that starts VSCode with the language server from the repository: `DAFNY_LANGUAGE_SERVER=./Binaries/DafnyLanguageServer.dll code`.
- Split the launch configuration "Run Extension" into "Run with default server" and "Run with $DAFNY_LANGUAGE_SERVER"
- Small update to `checkSupportedDotnetVersion` so it's agnostic to symlinks.

### Testing
Manually tested both features

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -21,7 +21,7 @@
       "preLaunchTask": "${defaultBuildTask}"
     },
     {
-      "name": "Run with custom server",
+      "name": "Run with $DAFNY_LANGUAGE_SERVER",
       "type": "extensionHost",
       "request": "launch",
       "args": [
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,7 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "Run Extension",
+      "name": "Run with default server",
       "type": "extensionHost",
       "request": "launch",
       "args": [
@@ -15,6 +15,9 @@
       "outFiles": [
         "${workspaceFolder}/dist/**/*.js"
       ],
+      "env": {
+        "DAFNY_LANGUAGE_SERVER": "",
+      },
       "preLaunchTask": "${defaultBuildTask}"
     },
     {
@@ -24,9 +27,6 @@
       "args": [
         "--extensionDevelopmentPath=${workspaceFolder}"
       ],
-      "env": {
-        "DAFNY_LANGUAGE_SERVER": "${env:CUSTOM_DAFNY_SERVER}", 
-      },
       "outFiles": [
         "${workspaceFolder}/dist/**/*.js"
       ],
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,6 +12,9 @@
       "args": [
         "--extensionDevelopmentPath=${workspaceFolder}"
       ],
+      "env": {
+        "DAFNY_LANGUAGE_SERVER": "${env:CUSTOM_DAFNY_SERVER}", 
+      },
       "outFiles": [
         "${workspaceFolder}/dist/**/*.js"
       ],
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,6 +12,18 @@
       "args": [
         "--extensionDevelopmentPath=${workspaceFolder}"
       ],
+      "outFiles": [
+        "${workspaceFolder}/dist/**/*.js"
+      ],
+      "preLaunchTask": "${defaultBuildTask}"
+    },
+    {
+      "name": "Run with custom server",
+      "type": "extensionHost",
+      "request": "launch",
+      "args": [
+        "--extensionDevelopmentPath=${workspaceFolder}"
+      ],
       "env": {
         "DAFNY_LANGUAGE_SERVER": "${env:CUSTOM_DAFNY_SERVER}", 
       },
--- a/src/dotnet.ts
+++ b/src/dotnet.ts
@@ -10,13 +10,12 @@ import { Messages } from './ui/messages';
 
 const ListRuntimesArg = '--list-runtimes';
 const execFileAsync = promisify(execFile);
-const lstatAsync = promisify(fs.lstat);
 
 // Returns an error message or undefined.
 export async function checkSupportedDotnetVersion(): Promise<string | undefined> {
   const { path: dotnetExecutable, manual } = await getDotnetExecutablePath();
   try {
-    const stats = await lstatAsync(dotnetExecutable);
+    const stats = await fs.promises.stat(dotnetExecutable);
     if(!stats.isFile()) {
       return dotnetExecutable + Messages.Dotnet.IsNotAnExecutableFile;
     }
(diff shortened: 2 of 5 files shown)


robin-aws:
YES, love it!

I'm a bit confused about why we need both environment variables though. Couldn't we have the `launch.json` pass through the same variable with something like `"DAFNY_LANGUAGE_SERVER": "${env:DAFNY_LANGUAGE_SERVER}"`?

src/dotnet.ts
```diff
@@ -10,14 +10,13 @@ import { Messages } from './ui/messages';
...
-const lstatAsync = promisify(fs.lstat);
 
 // Returns an error message or undefined.
 export async function checkSupportedDotnetVersion(): Promise<string | undefined> {
   const { path: dotnetExecutable, manual } = await getDotnetExecutablePath();
   try {
-    const stats = await lstatAsync(dotnetExecutable);
-    if(!stats.isFile() && !stats.isSymbolicLink()) {
+    const stats = await fs.promises.stat(dotnetExecutable);
+    if(!stats.isFile()) {
```
MikaelMayer, line 19: Did the previous code break something on your machine? Can you please elaborate on this change?

I know that if you don't add symlink check, then dotnet installations in /user/bin/dotnet which is a symlink would get undetected, as raised by @prvshah51 . So I really want to understand why you write a change that would break a configuration such as for @prvshah51 , what is the problem you are really trying to solve, and see how we can find some middle ground here.

src/dotnet.ts
```diff
@@ -10,14 +10,13 @@ import { Messages } from './ui/messages';
...
 const execFileAsync = promisify(execFile);
-const lstatAsync = promisify(fs.lstat);
 
 // Returns an error message or undefined.
 export async function checkSupportedDotnetVersion(): Promise<string | undefined> {
   const { path: dotnetExecutable, manual } = await getDotnetExecutablePath();
   try {
-    const stats = await lstatAsync(dotnetExecutable);
-    if(!stats.isFile() && !stats.isSymbolicLink()) {
+    const stats = await fs.promises.stat(dotnetExecutable);
```
MikaelMayer, line 18: Nice, I did not know the promises field of fs.

src/language/dafnyInstallation.ts
```diff
@@ -57,8 +57,10 @@ export function getLanguageServerRuntimePath(context: ExtensionContext): string
   return path.join(context.extensionPath, configuredPath);
 }
 
-function getConfiguredLanguageServerRuntimePath(): string | null {
-  return Configuration.get<string | null>(ConfigurationConstants.LanguageServer.RuntimePath);
+function getConfiguredLanguageServerRuntimePath(): string {
+  const languageServerOverride = process.env['DAFNY_LANGUAGE_SERVER'] ?? '';
+  const languageServerSetting = Configuration.get<string | null>(ConfigurationConstants.LanguageServer.RuntimePath) ?? '';
+  return languageServerOverride || languageServerSetting;
```
MikaelMayer, line 63: I'm surprised that you don't have a lint warning here, but, hey, that's good that it works, I like it!

robin-aws (approved):

.vscode/launch.json
```diff
@@ -6,7 +6,22 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "Run Extension",
+      "name": "Run with default server",
```
MikaelMayer, line 9: How do you ensure this configuration is the default one when we package the extension? I don't know if my question is relevant.

MikaelMayer (changes requested):
I like the idea of being able to configure the language server path from the environment temporarily, thanks for suggesting that !

I just wouldn't want developers to face the problem of changing the PATH variable globally once and then later not realizing that they cannot override it with the settings.json (everyone expects that a local setting overrides a global setting).

I would strongly suggest one or both of the two ideas to avoid this problem:
- Rename `DAFNY_LANGUAGE_SERVER` to `DAFNY_LANGUAGE_SERVER_OVERRIDE_PATH` to clearly indicate it's an override of even a local setting.
- In the Dafny output, add a message stating that it is starting Dafny from the environment path DAFNY_LANGUAGE_SERVER rather than the setting-set path, so if later someone has a DAFNY_LANGUAGE_SERVER defined globally and are going nuts because they are just starting VSCode which does not takes the setting DafnyLanguageServerPath into account anymore, they would immediately understand why.

Also, discussion about the removal of symlinks.

src/dotnet.ts
keyboardDrummer: Note that this code uses `stat` instead of `lstat`, so it works both with and without symlinks.

src/language/dafnyInstallation.ts
keyboardDrummer: That's because the type is `string` instead of `string | null`

src/dotnet.ts
MikaelMayer: Oh great ! Thanks for this finding. I did not know there was a difference between `stat` and `lstat`

.vscode/launch.json
keyboardDrummer: These configurations are only for development. The "main" method is configured in package.json/main and points to "./dist/extension.js"

src/language/dafnyInstallation.ts
```diff
@@ -57,8 +57,13 @@ export function getLanguageServerRuntimePath(context: ExtensionContext): string
   return path.join(context.extensionPath, configuredPath);
 }
 
-function getConfiguredLanguageServerRuntimePath(): string | null {
-  return Configuration.get<string | null>(ConfigurationConstants.LanguageServer.RuntimePath);
+function getConfiguredLanguageServerRuntimePath(): string {
+  const languageServerOverride = process.env['DAFNY_LANGUAGE_SERVER'] ?? '';
+  if(languageServerOverride) {
+    window.showInformationMessage('Using Dafny language server configured in $DAFNY_LANGUAGE_SERVER');
```
MikaelMayer, line 63: Since this is targeted to developers, I think we can 1) make the text shorter, and 2) use the space for giving the value of DAFNy_LANGUAGE_SERVER. I have two dafny folder on my machine, dafny and dafny2, so I would feel good getting a confirmation that it picked up the correct dafny folder.

```suggestion
    window.showInformationMessage(`Using $DAFNY_LANGUAGE_SERVER = ${languageServerOverride} for the server path`);
```

MikaelMayer (approved):
A minor suggestion, otherwise, everything looks good  ! Great job on notifying the user about the use of the language path.

cpitclaudel:
Don't you need a similar variable for where to find `Dafny.dll` (for compilation/execution?)
Maybe it should just be `DAFNY_DLLS=./Binaries/`?

keyboardDrummer:
> Don't you need a similar variable for where to find `Dafny.dll` (for compilation/execution?) Maybe it should just be `DAFNY_DLLS=./Binaries/`?

O, sorry I read this too late. I haven't run into this use-case yet but it sounds plausible. Since this is a development feature I think we're free to change the variable name at any point.

PR #173, opened 2022-06-08
Add DAFNY_DEV_SERVER variable

keyboardDrummer:
Allow configuring a `DAFNY_DEV_SERVER` env variable that's only used when starting the Dafny VSCode extension from source, but not when starting a regular VSCode instance.


cpitclaudel (approved):
Maybe it's worth documenting this variable in the extension's README?

keyboardDrummer:
> Maybe it's worth documenting this variable in the extension's README?

Will do in another PR

PR #175, opened 2022-06-10
Added missing status messages.

### LINKED ISSUE #174: Not knowing immediately if verification started again or not.
When I modify my code with verification on change, the status bar is not updated until the code actually is resolved or parsing or resolution failed.
This is misleading, because for large file, parsing and resolution can take a substantial time (a few seconds), during which it will still display the previous "Verifying... "message, which leads me to think that Dafny is stuck verifying since I changed the code and it does not update the status bar immediately.
We should reflect immediately that Dafny took the change into account and is processing it.

MikaelMayer:
Fixes #174 but will need Dafny Language Server to support those messages afterwards.

Alternative to discuss:
* Should we replace "Parsing started" with "Resolution started" and remove the second "Resolution started" message?

### DIFF
--- a/src/ui/compilationStatusView.ts
+++ b/src/ui/compilationStatusView.ts
@@ -9,12 +9,10 @@ const StatusBarPriority = 10;
 
 function toStatusMessage(status: CompilationStatus, message?: string | null): string {
   switch(status) {
-  case CompilationStatus.ParsingStarted:
-    return Messages.CompilationStatus.ParsingStarted;
-  case CompilationStatus.ParsingFailed:
-    return Messages.CompilationStatus.ParsingFailed;
   case CompilationStatus.ResolutionStarted:
     return Messages.CompilationStatus.ResolutionStarted;
+  case CompilationStatus.ParsingFailed:
+    return Messages.CompilationStatus.ParsingFailed;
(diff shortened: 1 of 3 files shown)


keyboardDrummer:
> Should we replace "Parsing started" with "Resolution started" and remove the second "Resolution started" message?

I think that's better. The user doesn't benefit from each having finished only parsing. It doesn't enable any additional IDE features.

MikaelMayer:
> I think that's better

Done. I also changed the message from "Resolution Started" to "Resolving..." which is more consistent with "Verifying..."

src/ui/compilationStatusView.ts
```diff
@@ -9,6 +9,8 @@ const StatusBarPriority = 10;
 
 function toStatusMessage(status: CompilationStatus, message?: string | null): string {
   switch(status) {
+  case CompilationStatus.ResolutionStarted:
```
keyboardDrummer, line 12: Shouldn't this be when it starts parsing already?

MikaelMayer: Yes, if resolution for the user is parsing + internal resolution, then resolution started is what we report.

PR #177, opened 2022-06-10
Fix syntax highlighting problem

### LINKED ISSUE #176: Wrong syntax highlighting for identifiers starting with "class"
image
image

MikaelMayer:
Fixes #176, the problem with identifiers starting with "class".
Manual test added.

### DIFF
--- a/syntaxes/Dafny.tmLanguage
+++ b/syntaxes/Dafny.tmLanguage
@@ -16,7 +16,7 @@
 	<array>
 		<dict>
 			<key>begin</key>
-			<string>(class)\s*([\w'?]+\b(?!&lt;))?\s*</string>
+			<string>(class\b)\s*([\w'?]+\b(?!&lt;))?\s*</string>
 			<key>beginCaptures</key>
 			<dict>
 				<key>1</key>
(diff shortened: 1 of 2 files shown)


syntaxes/Dafny.tmLanguage
```diff
@@ -16,7 +16,7 @@
 	<array>
 		<dict>
 			<key>begin</key>
-			<string>(class)\s*([\w'?]+\b(?!&lt;))?\s*</string>
+			<string>(class\b)\s*([\w'?]+\b(?!&lt;))?\s*</string>
```
cpitclaudel, line 19:
```suggestion
			<string>(class)\s+([\w'?]+\b(?!&lt;))?\s*</string>
```

cpitclaudel: Your solution is fine too, and I'm not sure whether `\s` allows a newline, but still putting this here.

MikaelMayer: I see what you mean, but that would prevent the parsing of `class{:extern}name` properly.

Issue #182, opened 2022-06-22
javac -version outputs to stderr


sorawee:
I am unable to install Dafny because the current installation code expects `javac -version` to output to stdout (as a way to test that Java is installed). However, in fact `javac -version` outputs to stderr, at least on my Mac M1.

sorawee:
This is also fixed by #183. Closing.

PR #183, opened 2022-06-22
Avoid downloading dmg when build from source & fix Java recognition

### LINKED ISSUE #180: Avoiding downloading dmg if it requires building from source anyway
We need to build from source when ["Found a non-supported architecture"](https://github.com/dafny-lang/ide-vscode/blob/master/src/language/dafnyInstallation.ts#L128). However, the installation process [downloads Dafny dmg regardless](https://github.com/dafny-lang/ide-vscode/blob/master/src/language/dafnyInstallation.ts#L122). It should avoid doing that.

sorawee:
Note that previously, Java recognition is done by testing `stdout`. I know for sure that this doesn't work on my system, as it should test `stderr` instead. However, I also do not know how other systems are. So I conservatively tested both, just to be safe.

Fixes #180, #182

cc: @MikaelMayer and @prvshah51

### DIFF
--- a/src/language/dafnyInstallation.ts
+++ b/src/language/dafnyInstallation.ts
@@ -166,8 +166,8 @@ export class DafnyInstaller {
     }
     try {
       const process = await this.execLog('javac -version');
-      if(!(/javac \d+\.\d+/.exec(process.stdout)) &&
-         !(/javac \d+\.\d+/.exec(process.stderr))) {
+      if(!(/javac \d+\.\d+/.exec(process.stdout))
+         && !(/javac \d+\.\d+/.exec(process.stderr))) {
         throw '';
       }
     } catch(error: unknown) {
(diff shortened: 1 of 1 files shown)


src/language/dafnyInstallation.ts
```diff
@@ -119,15 +119,15 @@ export class DafnyInstaller {
     this.writeStatus('Starting Dafny installation');
     try {
       await this.cleanInstallDir();
-      const archive = await this.downloadArchive(getDafnyDownloadAddress());
-      await this.extractArchive(archive);
-      await workspace.fs.delete(archive, { useTrash: false });
-      this.writeStatus('Dafny installation completed');
       if(os.type() === 'Darwin' && os.arch() !== 'x64') {
         // Need to build from source and move all files from Binary/ to the out/resource folder
         this.writeStatus(`Found a non-supported architecture OSX:${os.arch()}. Going to install from source and replace the automated installation.`);
```
MikaelMayer, line 124:
```suggestion
        this.writeStatus(`Found a non-supported architecture OSX:${os.arch()}. Going to install from source.`);
```

src/language/dafnyInstallation.ts
```diff
@@ -165,8 +165,9 @@ export class DafnyInstaller {
       return false;
     }
     try {
-      const result = (await this.execLog('javac -version')).stdout;
-      if(!(/javac \d+\.\d+/.exec(result))) {
+      const process = await this.execLog('javac -version');
+      if(!(/javac \d+\.\d+/.exec(process.stdout)) &&
+         !(/javac \d+\.\d+/.exec(process.stderr))) {
```
MikaelMayer, line 170: Nice fix !

MikaelMayer (changes requested):
Yes that should fix both problems, thank you very much for your contribution !
Just a rephrasing of the log.

PR #185, opened 2022-06-24
Fixed the test suite and added a test of the installer

MikaelMayer:
* I fixed very obsolete commands which prevented the test from running
* The test can now be run in VSCode by clicking on the launch configuration "Extension test"
* I added 3 mocking classes to simulate the workspace, the file system and the output channel.
* I added a test to verify what the installer says when it can't delete a previous extension.

The test is not yet running in CI. This will be for another PR.


src/test/suite/extension.test.ts
```diff
@@ -3,13 +3,54 @@ import * as assert from 'assert';
...
+  });
+
+  test('Installer checks', async () => {
+    const context = MockingUtils.mockedContext();
+    const m = MockingUtils.mockedOutputChannelBuilder();
+    const installer = new DafnyInstaller(context, m.outputChannel);
+    const execAsyncMock = MockingUtils.mockedExecAsync(MockingUtils.simpleCommandMap({
+      [''] : { stderr: '', stdout: '' }
+    }));
+    installer.execAsync = execAsyncMock;
```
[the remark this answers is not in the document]
MikaelMayer: I'll have to find how to convert the imports to require, but otherwise I think it's a neat idea.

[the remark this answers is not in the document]
MikaelMayer: I implemented your suggestion by adding a dev dependency to proxyquire, it looks like it's the exact library we will need to write more tests!

camrein:
So nice that you managed to fix this! 🚀

src/test/suite/extension.test.ts
```diff
@@ -1,12 +1,18 @@
...
+import * as proxyquire from 'proxyquire';
+import { MockingUtils, MockingExec } from './MockingUtils';
 
 // You can import and use all API from the 'vscode' module
 // as well as import your extension to test it
-import * as vscode from 'vscode';
+//import * as vscode from 'vscode';
+const vscode = require('vscode');
+const mockedExec = new MockingExec();
+proxyquire('child_process', {
```
keyboardDrummer, line 10: Shouldn't we use `proxyquire` to load `DafnyInstaller` ?
```javascript
const DafnyInstaller = proxyquire('DafnyInstaller', { 'child_process': { exec: mockedExec.stub } }
```

The current code makes it look like `child_process` is mutated for the remainder of the testing process.

src/test/suite/MockingUtils.ts
```diff
@@ -1,10 +1,30 @@
...
 import { ConfigurationConstants, LanguageServerConstants } from '../../constants';
-import { ExecAsyncType, ExecOutput } from '../../language/dafnyInstallation';
 import { MockedOutputChannelBuilder } from './MockedOutputChannelBuilder';
 import { MockedWorkspace } from './MockedWorkspace';
 
+export interface ExecOutput {
+  stdout: string;
+  stderr: string;
+}
+export class MockingExec {
```
keyboardDrummer, line 11: Have you looked at using sinon for creating the stubs? I was hoping we can create a stub based on the original type.

src/test/suite/extension.test.ts
MikaelMayer: Yes, thanks for your suggestion, I did not realize I could do this

However, my original code is intended. I don't want the code to actually run the original exec in any portion, so that I can test all exec commands. Hence, I mock the exec for the remaining of the file.
Every test can specify the behavior of exec that it expects.

src/test/suite/MockingUtils.ts
MikaelMayer: I just had a look at stubs, and it looks great, I'm just not sure about where to use it best.

In my first look, the version I found does not care about types being stubbed at all, it's just a wrapper around the JavaScript.
Also, I did not find a way it would support my use case for exec:

    public constructor() {
      this.stub = (command: string) => {
        throw `Unexpected comand ${command}`;
      };
    }

The best alternative I have found that sinon offers is nowhere as good:

    public constructor() {
      this.stub = sinon.stub().throwsArg()
    }

which throws the first argument. If I want to do anything more complicated than that, there are no other options. Writing a custom function is a better alternative to `sinon.stub()` in this case.
But that will only throw command, not a customized error message.

My conclusion is that, at this stage of development, I don't see a clear use case, so I would rather wait for the tests to become more complex, or working with someone else who has a clearer vision on how to use them.

src/test/suite/extension.test.ts
keyboardDrummer: > Hence, I mock the exec for the remaining of the file.

What makes you think this is just the remainder of the file? Wouldn't it mock exec for the entire test process including other files?

src/test/suite/extension.test.ts
```diff
@@ -1,15 +1,62 @@
...
 // as well as import your extension to test it
-import * as vscode from 'vscode';
-// import * as myExtension from '../../extension';
+//import * as vscode from 'vscode';
+const vscode = require('vscode');
+const mockedExec = new MockingExec();
+proxyquire('child_process', {
+  exec: mockedExec.stub
+});
+//import { ShowDocumentRequest } from 'vscode-languageclient';
```
keyboardDrummer, line 13: Can this be removed?

src/test/suite/extension.test.ts
MikaelMayer: I return the question to you. What makes you think it would mock exec for the entire test process? Proxyquire is local.

MikaelMayer: Ok, I found a way to make it import locally after running the document. I was probably wrong.

src/test/suite/extension.test.ts
```diff
@@ -1,15 +1,62 @@
 import * as assert from 'assert';
+import * as proxyquire from 'proxyquire';
+import { MockingUtils, MockingExec } from './MockingUtils';
 
 // You can import and use all API from the 'vscode' module
 // as well as import your extension to test it
-import * as vscode from 'vscode';
-// import * as myExtension from '../../extension';
+//import * as vscode from 'vscode';
+const vscode = require('vscode');
```
keyboardDrummer, line 8: Why use a require instead of an import?

keyboardDrummer (approved):

src/test/suite/extension.test.ts
MikaelMayer: To remember we can use proxyquire there down the road.

PR #186, opened 2022-06-24
Better dotnet and brew error messages

MikaelMayer:
* When dotnet is not installed or not found, the button meant to open the Dafny setting previously opened the command palette, which was completely pointless. Users would have had to type "Dafny" manually, and then navigate manually to the setting "Dotnet executable path". With this PR, this button now display immediately the option to the absolute path to the dotnet executable so that it can be modified straight away
* When brew fails to install dotnet, the error message is now much more explicit on the resolution steps, which I found consistently worked for mac M1 users.

### DIFF
--- a/src/language/dafnyInstallation.ts
+++ b/src/language/dafnyInstallation.ts
@@ -169,7 +169,7 @@ export class DafnyInstaller {
       > echo 'eval "$(/opt/homebrew/bin/brew shellenv)"' >> ~/.zprofile
       > eval "$(/opt/homebrew/bin/brew shellenv)"
 
-      and restart VSCode, possibly reinstalling Dafny.`);
+      and restart VSCode, which may reinstall Dafny.`);
       return false;
     }
     try {
--- a/src/language/dafnyInstallation.ts
+++ b/src/language/dafnyInstallation.ts
@@ -161,7 +161,15 @@ export class DafnyInstaller {
     try {
       await this.execLog('brew install dotnet-sdk');
     } catch(error: unknown) {
-      this.writeStatus('If you got `brew: command not found`, but brew is installed on your system, please add all brew commands to your ~/.zprofile, e.g. https://apple.stackexchange.com/a/430904 and reinstall Dafny.');
+      this.writeStatus('An error occurred while running this command.');
+      this.writeStatus(`${error}`);
+      this.writeStatus(`If brew is installed on your system, this can usually be resolved by adding add all brew commands to your ~/.zprofile,
+      e.g. by running the script there https://apple.stackexchange.com/a/430904 :
+
+      > echo 'eval "$(/opt/homebrew/bin/brew shellenv)"' >> ~/.zprofile
+      > eval "$(/opt/homebrew/bin/brew shellenv)"
+
+      and restart VSCode, possibly reinstalling Dafny.`);
       return false;
     }
     try {
--- a/src/startupCheck.ts
+++ b/src/startupCheck.ts
@@ -1,14 +1,15 @@
-import { window, commands, Uri } from 'vscode';
+import { window, commands, Uri, ExtensionContext, Extension } from 'vscode';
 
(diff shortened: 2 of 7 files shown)


prvshah51 (approved):

src/language/dafnyInstallation.ts
```diff
@@ -161,7 +161,15 @@ export class DafnyInstaller {
...
-      this.writeStatus('If you got `brew: command not found`, but brew is installed on your system, please add all brew commands to your ~/.zprofile, e.g. https://apple.stackexchange.com/a/430904 and reinstall Dafny.');
+      this.writeStatus('An error occurred while running this command.');
+      this.writeStatus(`${error}`);
+      this.writeStatus(`If brew is installed on your system, this can usually be resolved by adding add all brew commands to your ~/.zprofile,
+      e.g. by running the script there https://apple.stackexchange.com/a/430904 :
+
+      > echo 'eval "$(/opt/homebrew/bin/brew shellenv)"' >> ~/.zprofile
+      > eval "$(/opt/homebrew/bin/brew shellenv)"
+
+      and restart VSCode, possibly reinstalling Dafny.`);
```
keyboardDrummer, line 172: Consider changing `and restart VSCode, possibly reinstalling Dafny.` into `and restart VSCode, which may reinstall Dafny.`

src/startupCheck.ts
```diff
@@ -1,24 +1,27 @@
...
-async function checkDotnetInstallation(): Promise<boolean> {
+async function checkDotnetInstallation(context: ExtensionContext): Promise<boolean> {
   const answer = await checkSupportedDotnetVersion();
   if(answer !== undefined) {
     const selection = await window.showErrorMessage(
       answer,
       Messages.Dotnet.ChangeConfiguration,
       Messages.Dotnet.VisitDownload
     );
+    const extension = (context as unknown as { extension: Extension<void> }).extension;
```
keyboardDrummer, line 20: This looks like you're using a private API. Is there a way to do this using a public API?

src/language/dafnyInstallation.ts
MikaelMayer, line 172:
```suggestion
      and restart VSCode, which may reinstall Dafny.`);
```

src/startupCheck.ts
MikaelMayer: Good point.
This is actually documented public API, but for an unknown reason it's not exposed in VSCode.
https://code.visualstudio.com/api/references/vscode-api#ExtensionContext
It's not even written as "deprecated" and I did not find any alternative to access the extension's id (except maybe 'parsing' the URI, which isn't robust as in development mode the URI does not contain the ID)

src/language/dafnyInstallation.ts
MikaelMayer: Considered. It's more precise. Thanks !

src/startupCheck.ts
keyboardDrummer: I can write `context.extension.id` without issue.

keyboardDrummer (approved):

src/startupCheck.ts
```diff
@@ -1,14 +1,15 @@
-import { window, commands, Uri } from 'vscode';
+import { window, commands, Uri, ExtensionContext, Extension } from 'vscode';
```
MikaelMayer, line 1:
```suggestion
import { window, commands, Uri, ExtensionContext } from 'vscode';
```

src/startupCheck.ts
MikaelMayer: What experiment did you run so that you don't have issue? Here is my VSCode
image

And after fixing the warning in the CI, we get the same error:
> [tsl] ERROR in /home/runner/work/ide-vscode/ide-vscode/src/startupCheck.ts(23,25)
      TS2339: Property 'extension' does not exist on type 'ExtensionContext'.

Do you have any other idea about why this happens, or should I revert your commit?

Issue #188, opened 2022-06-28
"Show/Hide CounterExample" to toggle rather than be seperate


codyroux:
Rather than having 2 options `Show CounterExample` and `Hide CounterExample`, with different shortcuts, one could have a single option `Show/Hide CounterExample` which simply toggles the state (or even better, the toggle state would show up in the display, so that `Show CounterExample` would become `Hide CounterExample`).

Very minor quibble, but I'd also rather have it be "Counter Example" than "CounterExample", unless that's the exact name of the CLI option.

keyboardDrummer:
`counterexample` is a single word, so I'd suggest changing `CounterExample` to `Counterexample`

codyroux:
that's a fair point!

PR #191, opened 2022-06-29
Toogle counterexamples in context menu.

MikaelMayer:
Fixes both comments in #188

Feature
* The context menu shows either Show Counterexamples or Hide Counterexample, but not both

Benefits
* This enables to know whether Dafny is supposed to write counter-examples or not
* We remove one entry from the context menu.

Out of scope:
* I did not create a shortcut for toggling counter-examples, as advanced users might already rely on the F7 and F8 shortcuts and I did not want to replace those.
* I could have said that both "show counterexamples" and "hide counter examples" toggle the state, but it's a change of UX that I wanted to avoid for now.


README.md
```diff
@@ -8,7 +8,7 @@ This VSCode plugin requires the Dafny language server (shipped with the Dafny re
 - **Compile and Run** `.dfy` files.
 - **Verification** as one types.
 - **Syntax highlighting** thanks to [sublime-dafny](https://github.com/erggo/sublime-dafny). See file `LICENSE_sublime-dafny.rst` for license.
-- Display **CounterExample** for failing proof.
+- Display **Counterexample** for failing proof.
```
cpitclaudel, line 11:
```suggestion
- Display **counterexample** for failing proof.
```

README.md
```diff
@@ -8,7 +8,7 @@ This VSCode plugin requires the Dafny language server (shipped with the Dafny re
...
 - **IntelliSense** to suggest symbols.
 - **GoToDefinition** to quickly navigate.
```
cpitclaudel, line 13:
```suggestion
- **Go to definition** to quickly navigate.
```

package.json
```diff
@@ -75,12 +75,12 @@
         "when": "editorTextFocus && editorLangId == dafny"
       },
       {
-        "command": "dafny.showCounterExample",
+        "command": "dafny.showCounterexample",
```
cpitclaudel, line 78: I worry that changing command names could break things for clients.  Could it?

src/ui/counterExamplesView.ts
```diff
@@ -12,111 +12,120 @@ interface IColorOptions {
...
+    const contentText = Object.entries(counterexample.variables)
       .map(([ name, value ]) => `${name} = ${value}`)
       .join(' ');
     // TODO earlier versions showed a warning that there are references present.
-    const line = counterExample.position.line;
+    const line = counterexample.position.line;
     return {
       range: new Range(
-        new Position(line, counterExample.position.character + 1),
+        new Position(line, counterexample.position.character + 1),
```
cpitclaudel, line 128: Do you know where the typechecking warnings come from?

cpitclaudel (commented):
Code change looks great to me; I'm not sure about renaming commands, but happy to follow your lead there

README.md
```diff
@@ -45,7 +45,7 @@ When you first open a _Dafny_ file, the extension will prompt you to install .NE
...
 | `dafny.dotnetExecutablePath`      | Absolute path to the dotnet executable. Only necessary if dotnet is not in system PATH (you'll get an error if that's the case).                                                                   |                                                                |
-| `dafny.colorCounterExamples`      | Customize the color (HEX) of Counter Examples. There are two default colors: for dark theme (#0d47a1, #e3f2fd) and light theme (#bbdefb, #102027). This color setting will override both defaults. | `{ "backgroundColor": null, "fontColor": null }`               |
+| `dafny.colorCounterexamples`      | Customize the color (HEX) of Counter Examples. There are two default colors: for dark theme (#0d47a1, #e3f2fd) and light theme (#bbdefb, #102027). This color setting will override both defaults. | `{ "backgroundColor": null, "fontColor": null }`               |
```
MikaelMayer, line 48:
```suggestion
| `dafny.colorCounterexamples`      | Customize the color (HEX) of counterexamples. There are two default colors: for dark theme (#0d47a1, #e3f2fd) and light theme (#bbdefb, #102027). This color setting will override both defaults. | `{ "backgroundColor": null, "fontColor": null }`               |
```

package.json
MikaelMayer: I won't break anything for the extension itself, and the way to enter commands in VSCode via CTRL+SHIFT+P is case and space insensitive anyway, so I wouldn't worry about it.

cpitclaudel: Wouldn't it break custom shortcut configurations, for example? (I could be wrong here, I'm not familiar with vscode)

MikaelMayer: It might, I haven't tested that, good point. But I don't see why someone would assign more shortcuts than F7 and F8, so it probably concerns no one.

src/ui/counterExamplesView.ts
MikaelMayer: Yes. The newest version of VSCode considers positions to be `uinteger` instead of `number`. I added a cast.

package-lock.json
```diff
@@ -1,4737 +1,8 @@
 {
   "name": "ide-vscode",
   "version": "2.5.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
cpitclaudel, line 4: ?

cpitclaudel (approved):
LGTM.  Not sure about the bulk renaming, but if it's there it's there ^^  What's up with packageLock.json?

package-lock.json
MikaelMayer: I had to upgrade npm: `npm install -g npm@latest`

PR #196, opened 2022-07-07
Enable always using the latest nightly Dafny release

keyboardDrummer:
### Changes

Add an option "latest nightly" at the end of the "Preferred version" list, that will cause the IDE to download and use the latest nightly Dafny release

### Testing

Manually configured the IDE to use the latest nightly and used that, and the same for 3.7.


src/extension.ts
```diff
@@ -1,6 +1,7 @@
 import { Disposable, ExtensionContext, OutputChannel, window, commands } from 'vscode';
 import { ExtensionConstants, LanguageServerConstants } from './constants';
 import { DafnyCommands } from './commands';
+
```
MikaelMayer, line 4:
```suggestion
```

src/language/dafnyInstallation.ts
```diff
@@ -153,7 +175,7 @@ export class DafnyInstaller {
   }
 
   private async installFromSource() {
-    const installationPath = this.getCustomInstallationPath(os.arch());
+    const installationPath = await this.getCustomInstallationPath(os.arch());
```
MikaelMayer, line 178: There can now be a race between the custom installation path and the installation path (see below), because they both access a remote API at two different moments.
Could you please refactor the code as needed so that this API is invoked only once in this function?

Also, below at line 214, you have another call to
```
await this.execLog(`git checkout v${getConfiguredVersion()}`);
```
that you normally need to put an "await" in. But as per my comment above, if you ensure only one API call, you should be able to simplify it.
I'm looking forward the time when we will have tests to ensure such refactorings are not breaking the installation process.

MikaelMayer (changes requested):
I'm looking forward to use that extension more often !
Thank you very much for making nightly version available.
Just a question, do you know if it will re-trigger installation at every startup? Or do I have to reinstall the extension every day?

src/language/dafnyInstallation.ts
```diff
@@ -24,32 +25,53 @@ function ifNullOrEmpty(a: string | null, b: string): string {
...
+async function getConfiguredTagAndVersionUncached(): Promise<[string, string]> {
+  let version = Configuration.get<string>(ConfigurationConstants.PreferredVersion);
+  switch(version) {
+  case LanguageServerConstants.LatestStable:
+    version = LanguageServerConstants.LatestVersion;
+    break;
+  case LanguageServerConstants.LatestNightly: {
+    const result: any = await (await fetch('https://api.github.com/repos/dafny-lang/dafny/releases/tags/nightly')).json();
+    if(result.name !== undefined) {
+      const version = result.name.substring(6);
```
robin-aws, line 47:
```suggestion
      const version = result.name.substring("Dafny ".length);
```

Perhaps even with a sanity `startsWith` check beforehand :)

src/language/dafnyInstallation.ts
```diff
@@ -24,32 +25,53 @@ function ifNullOrEmpty(a: string | null, b: string): string {
...
+}
+
+async function getConfiguredTagAndVersionUncached(): Promise<[string, string]> {
+  let version = Configuration.get<string>(ConfigurationConstants.PreferredVersion);
+  switch(version) {
+  case LanguageServerConstants.LatestStable:
+    version = LanguageServerConstants.LatestVersion;
+    break;
+  case LanguageServerConstants.LatestNightly: {
+    const result: any = await (await fetch('https://api.github.com/repos/dafny-lang/dafny/releases/tags/nightly')).json();
```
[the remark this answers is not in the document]
keyboardDrummer: I would like the LSP server to be one of the commands in the Dafny CLI, so you would need to install the nightly build of the CLI dotnet tool. Maybe `dotnet tool search --prerelease dafny` can provide us with a list of versions, and then `dotnet tool install --tool-path=<something_containing_version> --version=<version>` can do the installation.

keyboardDrummer:
> Just a question, do you know if it will re-trigger installation at every startup? Or do I have to reinstall the extension every day?

It will retrigger

src/language/dafnyInstallation.ts
```diff
@@ -44,9 +50,14 @@ async function getConfiguredTagAndVersionUncached(): Promise<[string, string]> {
...
-      const version = result.name.substring(6);
-      return [ 'nightly', version ];
+      const name: string = result.name;
+      const versionPrefix = 'Dafny ';
+      if(name.startsWith(versionPrefix)) {
+        const version = name.substring(versionPrefix.length);
+        return [ 'nightly', version ];
+      }
     }
+    window.showWarningMessage('Failed to install latest nightly version of Dafny');
```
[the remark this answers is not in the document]
MikaelMayer: The path condition for this to happen is:

`result.name === undefined || !name.startsWith('Dafny ')`

so if result.name is undefined, we'll see `(got )` and if the name does not start with `Dafny `, we will see something like `(got dafny lowercase)`.

So not the whole result, only the name (which we could use to help troubleshooting).

src/language/dafnyInstallation.ts
```diff
@@ -24,32 +28,65 @@ function ifNullOrEmpty(a: string | null, b: string): string {
...
+  if(getConfiguredTagAndVersionCache === undefined) {
+    const result = await getConfiguredTagAndVersionUncached();
+    if(getConfiguredTagAndVersionCache === undefined) {
+      getConfiguredTagAndVersionCache = result;
+    }
+  }
+  return getConfiguredTagAndVersionCache;
+}
+
+async function getConfiguredTagAndVersionUncached(): Promise<[string, string]> {
```
MikaelMayer, line 47:
```suggestion
// Returns [git tagName, version displays name]
async function getConfiguredTagAndVersionUncached(): Promise<[string, string]> {
```

src/language/dafnyInstallation.ts
```diff
@@ -24,32 +28,65 @@ function ifNullOrEmpty(a: string | null, b: string): string {
...
+    const result: any = await (await fetch('https://api.github.com/repos/dafny-lang/dafny/releases/tags/nightly')).json();
+    if(result.name !== undefined) {
+      const name: string = result.name;
+      const versionPrefix = 'Dafny ';
+      if(name.startsWith(versionPrefix)) {
+        const version = name.substring(versionPrefix.length);
+        return [ 'nightly', version ];
+      }
+    }
+    window.showWarningMessage('Failed to install latest nightly version of Dafny. Using latest stable version instead.\n'
```
MikaelMayer, line 63:
```suggestion
    window.showWarningMessage('Failed to install latest nightly version of Dafny. Using latest version instead.\n'
```
We don't have "latest stable" in our versions, only "latest". Alternatively, we could rename "latest" into "latest stable"?

Also, the ordering that we are providing (latest, 3.7, 3.6, ... 3.2, latest nightly) might be a bit misleading because latest nightly is even more later than latest. But I think I understand that we don't want users to use latest nightly by default. So I don't have a better advice than what you did.

MikaelMayer (changes requested):
Tiny requested changes

src/language/dafnyInstallation.ts
keyboardDrummer: How about putting that information in the method name, then it doesn't go out of date so easily? 
Do you want to rename `getConfiguredTagAndVersionUncached` to `getConfiguredGitTagAndVersionUncached` ?

MikaelMayer: Ok, yes, the fact that the tag is a git tag was important to me. Funny enough, when I was looking at the signature, I was trying to understand what the return type was without looking at the function's name, so I guessed it based on the function's content.

PR #222, opened 2022-07-22
Added variable declaration to avoid mixing types and function call

### LINKED ISSUE #221: Syntax highlighting problem
```dafny
method Test() {
  var list: List<A> := new C<A>();
  if x == true { /** not a comment */
    print "problem";
  }
}
```
highlights like this:
image

The problem seems to come from the highlighter wrongly interpreting `List<A>` as the beginning of a function call (based on inspection of tokens and keys)

MikaelMayer:
Fixes #221 
Verified visually that all tests in the regression test still pass.

### DIFF
--- a/src/test/highlight/regression-highlight.dfy
+++ b/src/test/highlight/regression-highlight.dfy
@@ -1,3 +1,11 @@
+// Issue #221
+method Test() {
+  var list: List<A> := new C<A>();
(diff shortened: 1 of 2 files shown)


atomb (approved):
I think this looks reasonable, though I'm not an expert in TextMate language specifications.

PR #225, opened 2022-07-26
Set a default time limit of 20 seconds to avoid long freezes

cpitclaudel:
Closes https://github.com/dafny-lang/dafny/issues/1797


MikaelMayer (approved):
That's fair !

PR #227, opened 2022-07-26
chore: Clarify download prompt

### LINKED ISSUE #220: Installation without dotnet
The other day I was getting started on a fresh loaner Mac and tried installing Dafny through the extension. When given the download link I selected the wrong binary, not including ASP.NET Core Runtime 5.0 or 6.0. Can we include that information in the prompt before sending the user to select one unprepared?

fabiomadge:
fixes #220.


MikaelMayer (approved):
If it helps, let's do this. Thanks.

PR #229, opened 2022-07-27
Add verification management UI

keyboardDrummer:
### Changes

- Using the VSCode test API, provide an overview of available verification tasks, their running status, their outcome, and options to manually start and cancel verification task runs.
- Change the status bar implementation so the messages are also updated by manually triggered verification. Also the messages are slightly altered. When verification succeeds, if not everything was verified, then the number of skipped declarations is shown. When verification fails, the number of declarations that could not be proven is shown.
- Change verification gutter icons so they skip lines that have a test run icon from the VSCode testing API on them.

### Testing

- Manually tested the changes described above.
  - Test backwards compatibility mode
    - When using Dafny 3.7.3, the behavior is the same as it was before this PR.
    - When using the latest nightly and when using a custom LSP server, the new behavior activates .
  - Test that the new behavior works well when using
    - Never verify
    - Verify on change
    - Verify on save
  - Test with and without gutter icons enabled
- Will add automatic tests as well

### Screenshots

The following screenshots shows the effect of having "automatic verification" set to "never", having configured 2 cores for verification, and then manually clicking "run all" in the testing panel. First the two easy proofs m5 and fib verify quickly, then two slow proofs m3 and m4 are verifying, while m1 and m2 are still queued, and this moment is captured by the screenshot.
image

Once all the verification has completed, we see:
image

When using module and classes, the hierarchy is reflected in the test view:
image

Alternatively, a list view can be shown:
image


src/ui/verificationSymbolStatusView.ts
```diff
@@ -0,0 +1,318 @@
...
+  private automaticRunEnd: boolean = false;
+  private processUpdateLock: Promise<void> = Promise.resolve();
+
+  private createController(): TestController {
+    const controller = tests.createTestController('verificationStatus', 'Verification Status');
+    controller.createRunProfile('Verify', TestRunProfileKind.Run, async (request) => {
+      const items: TestItem[] = this.getItemsInRun(request, controller);
+      const runningItems: TestItem[] = [];
+      let outerResolve: () => void;
+      await this.processUpdateLock;
```
MikaelMayer, line 87: I think this will have the effect of letting the scheduler pick the next task before continuing. If this is intended, then you should name this `yieldAndRecoverLock()` as JavaScript is a single-threaded language.
Otherwise, can you please elaborate on why do you need this?

src/ui/verificationSymbolStatusView.ts
```diff
@@ -0,0 +1,318 @@
...
+  private readonly updateListenersPerFile: Map<string, ResolveablePromise<IVerificationSymbolStatusParams>> = new Map();
+  private readonly updatesPerFile: Map<string, IVerificationSymbolStatusParams> = new Map();
+  private readonly controller: TestController;
+  private automaticRunEnd: boolean = false;
+  private processUpdateLock: Promise<void> = Promise.resolve();
+
+  private createController(): TestController {
+    const controller = tests.createTestController('verificationStatus', 'Verification Status');
+    controller.createRunProfile('Verify', TestRunProfileKind.Run, async (request) => {
+      const items: TestItem[] = this.getItemsInRun(request, controller);
```
MikaelMayer, line 84:
```suggestion
      const testItems: TestItem[] = this.getItemsInRun(request, controller);
```

src/ui/verificationSymbolStatusView.ts
```diff
@@ -0,0 +1,318 @@
...
+      let outerResolve: () => void;
+      await this.processUpdateLock;
+      this.processUpdateLock = new Promise((resolve) => {
+        outerResolve = resolve;
+      });
+
+      const runs = items.map(item => this.languageClient.runVerification({ position: item.range!.start, textDocument: { uri: item.uri!.toString() } }));
+      for(const index in runs) {
+        const success = await runs[index];
+        if(success) {
```
MikaelMayer, line 95:
```suggestion
        const running = await runs[index];
        if(running) {
```

src/ui/verificationSymbolStatusView.ts
```diff
@@ -0,0 +1,318 @@
...
+    const result: TestItem[] = [];
+    const todo = run.include ? [ ...run.include ] : allItems;
+    const excludes = run.exclude ? new Set(run.exclude) : new Set<TestItem>();
+    while(todo.length > 0) {
+      const current = todo.pop()!;
+      if(excludes.has(current)) {
+        continue;
+      }
+
+      current.children.forEach(child => {
```
MikaelMayer, line 120: Is current.children ever not empty? Is the plan to nest the tests under modules, for example?

src/ui/verificationSymbolStatusView.ts
```diff
@@ -0,0 +1,318 @@
...
+    const document = await workspace.openTextDocument(uri.fsPath);
+    if(rootSymbols !== undefined) {
+
+      items = this.updateUsingSymbols(params, document, controller, rootSymbols);
+    } else {
+      items = params.namedVerifiables.map(f => this.getItem(document,
+        VerificationSymbolStatusView.convertRange(f.nameRange), controller, uri));
+    }
+
+    const runningItemsWithoutRun = params.namedVerifiables.
```
MikaelMayer, line 166: What do you mean here by "runningItemsWithoutRun"? Can you make it more explicit or add a comment?

src/ui/verificationSymbolStatusView.ts
```diff
@@ -0,0 +1,318 @@
...
+      map((element, index) => {
+        return { verifiable: element, testItem: items[index] };
+      }).
+      filter(({ verifiable, testItem }) => {
+        return this.itemStates.get(testItem.id) !== verifiable.status && this.itemRuns.get(testItem.id) === undefined;
+      }).map(r => r.testItem);
```
MikaelMayer, line 172: JavaScript: the language where periods indicate that the sentence is not finished and semicolon indicate that the sentence is finished. 🥂 
(No action needed for this comment)

src/ui/verificationSymbolStatusView.ts
```diff
@@ -0,0 +1,318 @@
...
+import { Messages } from './messages';
+
+interface ResolveablePromise<T> {
+  promise: Promise<T>;
+  resolve(value: T): void;
+}
+
+interface ItemRunState {
+  run: TestRun;
+  startedRunningTime?: number;
```
MikaelMayer, line 16: I think we had a consensus with @atomb @robin-aws @cpitclaudel and even @fabiomadge that displaying the running time for assertion batches is misleading, because there is already one method whose running time will be greater than all the other ones, for z3 to take the time to warm up. So we prefer resource unit.

However, here I think it's fine for now, because now the ExecutionEngine persists, this is for a top-level symbol, and the test API does not offer any other metric.
Just let's keep this in mind if later we want to extend this interface for single assertion batches.

src/ui/verificationSymbolStatusView.ts
```diff
@@ -0,0 +1,318 @@
...
+    const running = params.namedVerifiables.filter(v => v.status === PublishedVerificationStatus.Running);
+    const total = params.namedVerifiables.length;
+    let message: string;
+    if(running.length > 0) {
+      let verifying: string;
+      if(running.length === 1) {
+        const document = await workspace.openTextDocument(Uri.parse(params.uri));
+        verifying = document.getText(VerificationSymbolStatusView.convertRange(running[0].nameRange));
+      } else {
+        verifying = `${running.length} proofs`;
```
MikaelMayer, line 239: "proof" is a bit vague because there can be multiple proofs in a single method. I suggest we say we verified "declarations" or "definitions", which I think is more accurate.

```suggestion
        verifying = `${running.length} top-level declarations`;
```
or
```suggestion
        verifying = `${running.length} definitions`;
```
or
```suggestion
        verifying = `${running.length} declarations`;
```

src/ui/verificationSymbolStatusView.ts
```diff
@@ -0,0 +1,318 @@
...
+    } else {
+      const skipped = params.namedVerifiables.filter(v => v.status === PublishedVerificationStatus.Stale).length;
+      const errors = params.namedVerifiables.filter(v => v.status === PublishedVerificationStatus.Error).length;
+      const succeeded = completed - errors;
+
+      if(errors === 0) {
+        if(skipped === 0) {
+          message = Messages.CompilationStatus.VerificationSucceeded;
+        } else {
+          message = `Verified ${succeeded} proofs, skipped ${skipped}.`;
```
MikaelMayer, line 251:
```suggestion
          message = `Verified ${succeeded} declarations, skipped ${skipped}.`;
```

src/ui/verificationSymbolStatusView.ts
```diff
@@ -0,0 +1,318 @@
...
+        }
+      } else {
+        message = `${Messages.CompilationStatus.VerificationFailed} ${errors} proofs.`;
```
MikaelMayer, line 254:
```suggestion
        message = `${Messages.CompilationStatus.VerificationFailed} ${errors} declarations.`;
```

MikaelMayer (changes requested):
This is a great start ! A few comments and places where to improve.

* I want backwards compatibility with previous version of Dafny for now
* What if the test pane is closed?
  * I want the same behavior as before or very similar (have names of methods being currently verified in status bar)

src/ui/messages.ts
```diff
@@ -6,7 +6,7 @@ export namespace Messages {
     export const CompilationSucceeded = '$(book) Resolved (not verified)';
     export const Verifying = '$(sync~spin) Verifying';
     export const VerificationSucceeded = '$(thumbsup) Verification Succeeded';
-    export const VerificationFailed = '$(thumbsdown) Verification Failed';
+    export const VerificationFailed = '$(thumbsdown) Failed to verify';
```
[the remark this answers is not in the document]
keyboardDrummer: The intended change here is to also display how many proofs failed to verify, so instead of "Verification Failed" it says "Failed to verify X proofs". There's no real need for this change but I thought it would be a nice extra bit of information since that is available now.

"Could not prove X proofs" also seems fine to me.

src/language/api/verificationSymbolStatusParams.ts
```diff
@@ -0,0 +1,19 @@
...
+  uri: DocumentUri;
+  version?: integer;
+  namedVerifiables: NamedVerifiableStatus[];
+}
+export interface NamedVerifiableStatus {
+  nameRange: Range;
+  status: PublishedVerificationStatus;
+}
+
+export enum PublishedVerificationStatus {
```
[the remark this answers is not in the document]
keyboardDrummer: A link to `https://github.com/dafny-lang/dafny/blob/afd3138abffec0ddeace3f0d79bd7535290f39fd/Source/DafnyLanguageServer/Workspace/FileVerificationStatus.cs#L45` ?

src/ui/compilationStatusView.ts
```diff
@@ -19,21 +19,12 @@ function toStatusMessage(status: CompilationStatus, message?: string | null): st
...
-    return message != null
-      ? `${Messages.CompilationStatus.Verifying} ${message}...`
-      : `${Messages.CompilationStatus.Verifying}...`;
-  case CompilationStatus.VerificationSucceeded:
-    return Messages.CompilationStatus.VerificationSucceeded;
-  case CompilationStatus.VerificationFailed:
-    return message != null
-      ? `${Messages.CompilationStatus.VerificationFailed} ${message}`
-      : `${Messages.CompilationStatus.VerificationFailed}`;
+  default: throw new Error(`Should not handle status message ${status}`);
```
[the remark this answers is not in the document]
keyboardDrummer: There's a change before this method is called that prevents calling it for these types of messages.

src/ui/compilationStatusView.ts
```diff
@@ -52,8 +43,6 @@ export default class CompilationStatusView {
     context.subscriptions.push(
       commands.registerCommand(DafnyCommands.OpenStatusBarMenu, () => statusBarActionView.openStatusBarMenu()),
       languageClient.onCompilationStatus(params => view.compilationStatusChanged(params)),
-      languageClient.onVerificationStarted(params => view.verificationStarted(params)),
-      languageClient.onVerificationCompleted(params => view.verificationCompleted(params)),
```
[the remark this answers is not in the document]
keyboardDrummer: So.. we should do a version check here to see which LSP server version we're using, and if we're using an old one still execute these lines?

keyboardDrummer:
> * I want backwards compatibility with previous version of Dafny for now
So we want the client to be backwards compatible, so it can work with old versions of Dafny, that makes sense to me.

For the server, do we also want to keep it backwards compatible with old IDEs? Seems like that's not entirely necessary.

> * What if the test pane is closed?
>   * I want the same behavior as before or very similar (have names of methods being currently verified in status bar)

Let's discuss synchronously

src/ui/verificationGutterStatusView.ts
```diff
@@ -254,23 +258,31 @@ export default class VerificationGutterStatusView {
 
   // Converts the IVerificationStatusGutter to a map from line verification status
   // to an array of ranges that VSCode can consume.
-  private getRangesOfLineStatus(params: IVerificationGutterStatusParams): Map<LineVerificationStatus, Range[]> {
+  private async getRangesOfLineStatus(params: IVerificationGutterStatusParams): Promise<Map<LineVerificationStatus, Range[]>> {
     const perLineStatus = this.addCosmetics(params.perLineStatus);
 
-    let previousLineStatus = -1;
-    let initialStatusLine = -1;
+    const symbolParams = await this.symbolStatusView.getFirstStatusForCurrentVersion(params.uri);
```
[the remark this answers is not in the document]
keyboardDrummer: I don't think there is a delay, because symbol statuses (and gutter messages?) are published immediately after translation.

src/language/dafnyInstallation.ts
```diff
@@ -28,6 +28,27 @@ async function ifNullOrEmpty(a: string | null, b: () => Promise<string>): Promis
...
+  const [ , version ] = await getConfiguredTagAndVersion();
+  if(!/(\d)\.(\d)\.(\d)/g.test(version)) {
+    return undefined;
+  }
+  return version;
+}
+
+export function versionToNumeric(version: string | undefined): number {
+  if(version === undefined) {
+    return Number.MAX_VALUE;
```
MikaelMayer, line 46: Aha nice hack to ensure a custom installation path is treated as the most recent Dafny version.
However, this might not always be true: If I checkout a previous installation of Dafny to test a certain behavior, it will treat it like it was the last version and your code will fail to display previous messages.
I should have thought of that earlier, but instead of checking the server's path and inferred version, could you please look into "capabilities"?
I see a message that the language server sends to VSCode at the very beginning:

```
Received response 'initialize - (0)' in 2312ms.
Result: {
    "capabilities": {
        "experimental": {},
        "workspace": {
            "workspaceFolders": {
                "supported": true,
                "changeNotifications": true
            },
            "fileOperations": {}
        }
    },
    "serverInfo": {
        "name": "DafnyLanguageServer"
    }
```

I think you should be able to detect there if the language server has the capability to send the test items. That will be more robust than testing the version (testing the version would be as robust but only if the serverInfo contained the version number of Dafny...)

src/language/dafnyLanguageClient.ts
```diff
@@ -142,6 +142,16 @@ export class DafnyLanguageClient extends LanguageClient {
     this.diagnosticsListeners.push(callback);
   }
 
+  // Removed as of Dafny 3.8
```
MikaelMayer, line 145:
```suggestion
  // Backward compatibility for versions of Dafny <= 3.2.X
```

The removal of this API was added on July 29, 2021, before Dafny 3.3.0 was released 
https://github.com/dafny-lang/dafny/commit/5df5c01decd1ebd523f0f644e5b3af0789f6728d
(I found it using `git log -S dafny/verification/started`)

Can you fix the other one below as well, and also mention in `compilationStatusView:91` where it says `Legacy methods` to state explicitly what version of Dafny this is used for? That way, it will be easier to decide when we drop compatibility (i.e. during a major bump of the extension's version)

src/ui/compilationStatusView.ts
```diff
@@ -19,7 +19,20 @@ function toStatusMessage(status: CompilationStatus): string {
     return Messages.CompilationStatus.ResolutionFailed;
   case CompilationStatus.CompilationSucceeded:
     return Messages.CompilationStatus.CompilationSucceeded;
-  default: throw new Error(`Should not handle status message ${status}`);
+
+  // Not used from 3.8 on
```
MikaelMayer, line 23:
```suggestion
  // Backward compatibility for versions <= 3.7.X
```
Alternatively, we haven't decided the numbering of the new version of Dafny, whether it'll be 3.8 or 3.7.4. so for now I'd only mention the Dafny commit and date until we get a proper release.
Whatever you choose, please also replace the messages `Not used from 3.8 on` below.

src/ui/compilationStatusView.ts
```diff
@@ -52,24 +66,52 @@ export default class CompilationStatusView {
...
+
+
   private static readonly handledMessages = new Set([
     CompilationStatus.ResolutionStarted,
     CompilationStatus.ParsingFailed,
     CompilationStatus.ResolutionFailed,
     CompilationStatus.CompilationSucceeded ]);
 
-  private compilationStatusChanged(params: ICompilationStatusParams): void {
+  public compilationStatusChanged38(params: ICompilationStatusParams): void {
```
MikaelMayer, line 114: Did you mean?
```suggestion
  public compilationStatusChanged37(params: ICompilationStatusParams): void {
```
I understand this is the old method that is going to be deprecated, so it should contain the version number that it is useful for.

src/ui/statusBarActionView.ts
```diff
@@ -76,7 +76,8 @@ export default class StatusBarActionView {
       RestartDafny
     ];
     const currentDocumentStatus = this.compilationStatusView.getCurrentDocumentStatus();
-    if(currentDocumentStatus === Messages.CompilationStatus.VerificationFailed
+    if(currentDocumentStatus === Messages.CompilationStatus.VerificationFailedOld
```
MikaelMayer, line 79: Good job on making this backward compatible. I wonder if there is a better way to not depend on these messages. No action needed here if you don't have an idea.

src/ui/messages.ts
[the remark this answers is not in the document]
MikaelMayer: `Could not prove X declarations` looks good to me, thanks

src/language/dafnyInstallation.ts
keyboardDrummer: Capabilities seem useful when there's a protocol and different servers implement different parts of it. For Dafny, we have Dafny specific protocol extensions and the latest Dafny server will always implement everything in the latest protocol version, so I think a simple API version number should be sufficient.

But I agree it would be best if the server vends this API version number.

src/ui/verificationGutterStatusView.ts
```diff
@@ -261,28 +261,28 @@ export default class VerificationGutterStatusView {
   private async getRangesOfLineStatus(params: IVerificationGutterStatusParams): Promise<Map<LineVerificationStatus, Range[]>> {
     const perLineStatus = this.addCosmetics(params.perLineStatus);
 
-    const symbolParams = await this.symbolStatusView.getFirstStatusForCurrentVersion(params.uri);
-    const linesToSkip = new Set(symbolParams.namedVerifiables.map(v => v.nameRange.start.line));
+    const symbolParams = (await this.symbolStatusView?.getFirstStatusForCurrentVersion(params.uri)) ?? [];
+    const linesToSkip = [ -1, ...symbolParams.map(testItem => testItem.range!.start.line).sort((a, b) => a - b), perLineStatus.length ];
```
MikaelMayer, line 265: Since you are changing substantially this method. could you please
1) Refactor the rest of the code after this line in a public static helper method named `PerLineStatusToRanges`, taking `perLineStatus` and `linesToSkip`  as arguments (and prepend `-1` to `lineToSkip`)
2) At the end of `src/test/extensions.test.ts`, add the following

```javascript
suite("Verification Gutter", () => {
  test("PerLineStatusToRanges", () => {
    var ranges= VerificationGutterStatusView.PerLineStatusToRanges([1, 1, 1, 3, 3, 3, 2, 2, 2], [2, 4, 6]);
    assert.strictEqual(3, result.size);
    assert.deepStrictEqual([new Range(0, 1, 1, 1)], ranges.get(1));
    assert.deepStrictEqual([new Range(7, 1, 7, 1)], ranges.get(2));
    assert.deepStrictEqual([new Range(3, 1, 3, 1), new Range(5, 1, 5, 1)], ranges.get(3));
  })
});
```

3) Run the tests to verify that it works.

src/ui/verificationSymbolStatusView.ts
```diff
@@ -62,17 +71,35 @@ export default class VerificationSymbolStatusView {
...
+  public constructor(
     private readonly context: ExtensionContext,
     private readonly languageClient: DafnyLanguageClient,
     private readonly compilationStatusView: CompilationStatusView) {
     this.controller = this.createController();
+
+    workspace.onDidChangeTextDocument(e => {
+      const uriString = e.document.uri.toString();
+      this.updatesPerFile.delete(uriString);
+      this.updateListenersPerFile.delete(uriString);
```
MikaelMayer, line 83: Will these updates be automatically immediately regenerated? Couldn't you delete them only when the language server returns new ones to override?

src/ui/verificationSymbolStatusView.ts
```diff
@@ -231,14 +266,9 @@ export default class VerificationSymbolStatusView {
...
-      if(running.length === 1) {
-        const document = await workspace.openTextDocument(Uri.parse(params.uri));
-        verifying = document.getText(VerificationSymbolStatusView.convertRange(running[0].nameRange));
-      } else {
-        verifying = `${running.length} proofs`;
-      }
-      message = `${Messages.CompilationStatus.Verifying} ${verifying}, completed ${completed}/${total}`;
+      const document = await workspace.openTextDocument(Uri.parse(params.uri));
+      const verifying = running.map(item => document.getText(VerificationSymbolStatusView.convertRange(item.nameRange))).join(', ');
+      message = `$(sync~spin) Verified ${completed}/${total}, verifying ${verifying}`;
```
MikaelMayer, line 271: I liked this message which is much more precise than the previous one.

MikaelMayer (changes requested):
It's taking a great shape !
There was at least one unanswered question from my first review as well.

src/ui/compilationStatusView.ts
keyboardDrummer: Cutting a Dafny PR to increment the minor version number so it'll be 3.8.0 at minimum

src/ui/compilationStatusView.ts
keyboardDrummer: To be consistent with the comments "// Backwards compatibility for versions of Dafny <= 3.8.0" I think it's better to mention 3.8, but I've made the function name more explicit "compilationStatusChangedForBefore38"

src/test/suite/extension.test.ts
```diff
@@ -112,13 +112,11 @@ suite('Dafny IDE Extension Installation', () => {
...
-
-  test('Sample test', () => {
-    assert.strictEqual(-1, [ 1, 2, 3 ].indexOf(5));
-    assert.strictEqual(1, [ 1, 2, 3 ].indexOf(2));
+suite('Verification Gutter', () => {
+  test('perLineStatusToRanges', () => {
+    const ranges = VerificationGutterStatusView.perLineStatusToRanges([ 1, 1, 1, 0, 0, 0, 2, 2, 2 ], [ 2, 4, 6 ]);
+    assert.deepStrictEqual([ new vscode.Range(0, 1, 1, 1) ], ranges.get(1));
+    assert.deepStrictEqual([ new vscode.Range(7, 1, 8, 1) ], ranges.get(2));
+    assert.deepStrictEqual([ new vscode.Range(3, 1, 3, 1), new vscode.Range(5, 1, 5, 1) ], ranges.get(0));
```
MikaelMayer, line 120: Tests passing ! Fantastic !!

src/ui/dafnyIntegration.ts
```diff
@@ -20,9 +19,8 @@ export default async function createAndRegisterDafnyIntegration(
   GhostDiagnosticsView.createAndRegister(context, languageClient);
   const compilationStatusView = CompilationStatusView.createAndRegister(context, languageClient, languageServerVersion);
   let symbolStatusView: VerificationSymbolStatusView | undefined = undefined;
-  const apiVersion = await dafnyInstallation.getServerApiVersion();
-  const usingDafny38OrHigher = dafnyInstallation.versionToNumeric('3.8.0') <= dafnyInstallation.versionToNumeric(apiVersion);
-  if(usingDafny38OrHigher) {
+  const serverSupportsSymbolStatusView = versionToNumeric('3.8.0') <= versionToNumeric(languageServerVersion);
```
MikaelMayer, line 22: Oh that was that simple?! Great !

src/ui/verificationSymbolStatusView.ts
MikaelMayer: This comment as well.

MikaelMayer (approved):
Great job, can't wait to use it !

PR #231, opened 2022-07-29
Syntax highlighting after "var" fixed

### LINKED ISSUE #230: Syntax highlighter
The following fails syntax highlighting
```dafny
module A {
  trait T {
    var x : int
    var y: int
  }
}
```
image
Notably:
* T is not highlighted as a type
* Everything after the first "var" is just not highlighted anymore.

It looks like "var" starts to be highlighted both as a keyword and an identifier, and then everything else after it is considered part of the identifier.

MikaelMayer:
The issue is that the syntax highlighter was expecting "var" to be followed by a semicolon ";". If not, it would never exit the fact that it's a "var" declaration and would not highlight anything else.
I added a few keywords and ensured that the type after the var is highlighted as a type as well.
Now:

image

Fixes #230

### DIFF
--- a/syntaxes/Dafny.tmLanguage
+++ b/syntaxes/Dafny.tmLanguage
@@ -26,15 +26,7 @@
 				</dict>
 			</dict>
 			<key>end</key>
-			<string>:=|:\||;|(?=const|least|greatest|predicate|function|method|lemma|var|type|datatype|newtype)</string>
-			<!--key>endCaptures</key>
-			<dict>
-				<key>1</key>
-				<dict>
-					<key>name</key>
-					<string>storage.type.dafny</string>
-				</dict>
-			</dict-->
+			<string>:=|:\||;|\)|\}|(?=(?:class|codatatype|const|constructor|datatype|function|greatest|iterator|least|lemma|method|module|newtype|predicate|trait|type|var)\b)</string>
 			<key>name</key>
 			<string>meta.vardecl.dafny</string>
 			<key>patterns</key>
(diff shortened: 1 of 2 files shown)


syntaxes/Dafny.tmLanguage
```diff
@@ -26,17 +26,17 @@
 				</dict>
 			</dict>
 			<key>end</key>
-			<string>:=|:\||;</string>
-			<key>endCaptures</key>
+			<string>:=|:\||;|(?=const|least|greatest|predicate|function|method|lemma|var|type|datatype|newtype)</string>
```
cpitclaudel, line 29: There are entries missing in this list.  Here's the one that dafny-mode uses:

```
(defconst dafny-defuns
  '("class" "codatatype" "const" "constructor" "datatype" "function"
    "iterator" "lemma" "method" "newtype" "predicate" "trait" "type"
    "function method" "predicate method" "module"
    "least predicate" "greatest predicate" "least lemma" "greatest lemma"))
```

cpitclaudel (changes requested):

syntaxes/Dafny.tmLanguage
cpitclaudel: But then why does the regex have `datatype`, `type`, and `newtype`?

MikaelMayer: Good point. I was going to say, one never knows if these won't be part of a class one day but ... nah ok let me add all of them, it's not costlier

cpitclaudel (approved):

PR #232, opened 2022-08-01
Add missing subscription so restart server logic works again

keyboardDrummer:


MikaelMayer (approved):
Well spotted ! I had wondered why restarting the server wasn't working reliably. How did you figure out this missing subscription?

keyboardDrummer:
> Well spotted ! I had wondered why restarting the server wasn't working reliably. How did you figure out this missing subscription?

Just from the exception that complained about a duplicate registration for a particular name :)

PR #234, opened 2022-08-02
Add verification symbols test

keyboardDrummer:
Fixes https://github.com/dafny-lang/ide-vscode/issues/236

### Changes
- Run VSCode backed integration tests using the nightly Dafny language server
- Always use a mocked `vscode` in the existing tests in `src/test/suite/extension.test.ts`. Note the missing `@global: true` annotation was added in the proxyquire usage.
- Fix a bug that prevented verification management from working with unsaved files
- Fix a bug that broke the restart server feature

### Testing
- [WIP] Add a test for the `dafny.restartServer` command. I had to turn this test off because it causes the following test to fail. It triggers an issue described here: https://github.com/microsoft/vscode-languageserver-node/issues/755
- Add a integration test for verification management


src/test/suite/extension.test.ts
```diff
@@ -119,4 +122,23 @@ suite('Verification Gutter', () => {
...
+method Foo(x: nat) returns (y: nat) 
+  ensures y > 2;
+{ 
+  return x + 2; 
+}`;
+    const extension = vscode.extensions.getExtension('dafny-lang.ide-vscode')!;
+    await extension.activate();
+    const document = await vscode.workspace.openTextDocument({ language: 'dafny', content: program });
+    const symbols1 = await vscode.commands.executeCommand('vscode.executeDocumentSymbolProvider', document.uri) as DocumentSymbol[];
+    await vscode.commands.executeCommand('dafny.restartServer');
```
MikaelMayer, line 139: Why do you need to restart the server here? Are you only checking if the server returns the same commands?

MikaelMayer (commented):
It seems valuable, but why is the test "skip" then?

.vscode/launch.json
```diff
@@ -39,9 +39,11 @@
       "request": "launch",
       "runtimeExecutable": "${execPath}",
       "args": [
+        "--disable-extensions",
```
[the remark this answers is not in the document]
keyboardDrummer: Just the others

src/test/suite/extension.test.ts
keyboardDrummer: This test is for testing the restartServer command, so it should execute it right? It was correctly identifying the bug that this PR fixes, but sadly it also causes the subsequent tests to fail, so I can't merge it for now.

keyboardDrummer:
> It seems valuable, but why is the test "skip" then?

I can also remove it, but I haven't been able to get it to work without breaking other tests.

MikaelMayer (approved):
For now, I'm approving it, expecting we find a way to make the test pass later.

PR #244, opened 2022-08-04
Added missing awaits in publish release process

MikaelMayer:


robin-aws (approved):
LGTM - how much work would it be to set up nightly releases for this as well? Great way to catch typos. :)

PR #249, opened 2022-08-10
Correctly handle a difference between verifiable and document symbols

keyboardDrummer:
### Changes

Correctly handle a difference between verifiable symbols and document symbols

### Testing

Added a unit test

### DIFF
--- a/src/ui/verificationSymbolStatusView.ts
+++ b/src/ui/verificationSymbolStatusView.ts
@@ -292,7 +292,6 @@ export default class VerificationSymbolStatusView {
       childCollection.replace(newChildren);
     }
 
-    this.getItemsFilePromise(params.uri).resolve([ ...itemMapping.values() ]);
     return items;
   }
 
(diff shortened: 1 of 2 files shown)


src/ui/verificationSymbolStatusView.ts
```diff
@@ -173,7 +173,7 @@ export default class VerificationSymbolStatusView {
     if(rootSymbols !== undefined) {
       items = this.updateUsingSymbols(params, document, controller, rootSymbols);
     } else {
-      items = params.namedVerifiables.map(f => this.getItem(document,
+      items = params.namedVerifiables.map(f => VerificationSymbolStatusView.getItem(document,
```
MikaelMayer, line 176: Here it's where you solve this bug, correct? You are taking the most up-to-date collection instead of the items in the class that might not be up-to-date?

keyboardDrummer: This is just an instance method that became static. The bugfix is in the changes to `updateUsingSymbols`

MikaelMayer (approved):

PR #252, opened 2022-08-16
Prevent overlap of play and gutter icons on parent nodes

keyboardDrummer:
### Changes

- Prevent overlap of play and gutter icons on parent nodes such as classes and modules.
- Prevent visible and internal state of testItems from getting out of sync, which could occur when cancelling a run while it was returning results.

### Testing

Manually tested

### DIFF
--- a/src/ui/verificationSymbolStatusView.ts
+++ b/src/ui/verificationSymbolStatusView.ts
@@ -167,22 +167,30 @@ export default class VerificationSymbolStatusView {
     document: TextDocument,
     rootSymbols: DocumentSymbol[] | undefined) {
 
-    this.getVerifiableRangesPromise(params.uri).resolve(params.namedVerifiables.map(v => VerificationSymbolStatusView.convertRange(v.nameRange)));
     this.updateStatusBar(params);
     this.updatesPerFile.set(params.uri, params);
     const controller = this.controller;
-    let items: TestItem[];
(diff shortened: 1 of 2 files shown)


src/ui/verificationSymbolStatusView.ts
```diff
@@ -167,33 +167,43 @@ export default class VerificationSymbolStatusView {
...
+      controller.items.replace(leafItems);
     }
+    function collectTestItems(collection: TestItemCollection, result: TestItem[]) {
+      collection.forEach(item => {
+        result.push(item);
+        collectTestItems(item.children, result);
+      });
+    }
+    const allTestItems: TestItem[] = [];
+    collectTestItems(controller.items, allTestItems);
```
MikaelMayer, line 188:
```suggestion
    const allTestItems: TestItem[] = [];
    function collectTestItems(collection: TestItemCollection) {
      collection.forEach(item => {
        allTestItems.push(item);
        collectTestItems(item.children);
      });
    }
    collectTestItems(controller.items);
```
It's more legible with one less argument, unless you plan to lift this function outside of this scope?

MikaelMayer (commented):
Looks great ! I have on small suggestion, otherwise I'll approve it.

MikaelMayer (approved):
Great ! Thank you very much.

PR #253, opened 2022-08-18
Do not show verified status message when no solvers are free

keyboardDrummer:


MikaelMayer (approved):
That will surely explain better why the interface is sometimes frozen.

PR #254, opened 2022-08-19
Do not show status updates for non-active documents

keyboardDrummer:


MikaelMayer (approved):
This seems very useful.

PR #258, opened 2022-08-29
Fix highlighting of labelled statements

### LINKED ISSUE #240: Label syntax highlighting missing
It looks like the syntax highlighter completely forgot about labelled statements.
For example:

```dafny
method Test(someVar: int)
  requires zero: someVar > 0
{
    label before: otherMethod();
    assert notzero: someVar != 0 by {
        reveal zero;
    }
}

method otherMethod() {
}
```

is rendered as:
image

Note how in the requires, label and assert statement, the token after the label is misinterpreted as a type.

MikaelMayer:
This PR fixes #240 by checking that the ":" is not preceeded by "label <ident>", "assert <indent>" and "requires <indent>", so that its right-hand side is not misinterpreted as a type.

### DIFF
--- a/src/test/highlight/regression-highlight.dfy
+++ b/src/test/highlight/regression-highlight.dfy
@@ -1,3 +1,19 @@
+// Issue #240
+method Test(someVar: int)
+  requires zero: someVar > 0
+              // ^^^^^^^ should not look like a type (the int above)
+{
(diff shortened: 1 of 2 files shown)


prvshah51:
I have tested it and it's working.

prvshah51 (approved):

PR #260, opened 2022-08-30
Fix disappearing resolution errors with verification tests.

### LINKED ISSUE #259: New test extension does not display parser and resolution errors anymore in the gutter
If deactivate the option "verification as tests"
image

If activate the option "verification as tests"
image

The icon is no longer displayed and the gutter is no longer grayed out.
The problem comes from the extension as the server correctly sends out the correct gutter icons in both case.

MikaelMayer:
Fixes #259 by short-circuiting the line skipping if there are resolution errors.

### DIFF
--- a/src/ui/verificationGutterStatusView.ts
+++ b/src/ui/verificationGutterStatusView.ts
@@ -261,7 +261,9 @@ export default class VerificationGutterStatusView {
   private async getRangesOfLineStatus(params: IVerificationGutterStatusParams): Promise<Map<LineVerificationStatus, Range[]>> {
     const perLineStatus = this.addCosmetics(params.perLineStatus);
 
-    const symbolParams = (await this.symbolStatusView?.getVerifiableRanges(params.uri)) ?? [];
+    const symbolParams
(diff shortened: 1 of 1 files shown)


src/ui/verificationGutterStatusView.ts
```diff
@@ -261,7 +261,9 @@ export default class VerificationGutterStatusView {
   private async getRangesOfLineStatus(params: IVerificationGutterStatusParams): Promise<Map<LineVerificationStatus, Range[]>> {
     const perLineStatus = this.addCosmetics(params.perLineStatus);
 
-    const symbolParams = (await this.symbolStatusView?.getVerifiableRanges(params.uri)) ?? [];
+    const symbolParams
+      = params.perLineStatus.indexOf(LineVerificationStatus.ResolutionError) > 0 ? []
+        : (await this.symbolStatusView?.getVerifiableRanges(params.uri)) ?? [];
```
keyboardDrummer, line 266: Could you also change this to `await (this.symbolStatusView?.getVerifiableRanges(params.uri) ?? Promise.resolve([]))` ?

keyboardDrummer (approved):

PR #261, opened 2022-09-01
Enable status bar also when verification as tests is turned off

keyboardDrummer:


MikaelMayer (approved):
Very good, I see the problem we had before and now it's correctly fixed.
Any change you can add some tests?

PR #270, opened 2022-09-20
Fix: Number followed by `<` incorrectly highlighted as a generic

### LINKED ISSUE #269: Syntax highlighting bug in BinaryExpr
I noticed a bug in the syntax highlighting of VSCode extension that might be from DafnyAST but I am not sure so I am reporting it here.
In the following example, adding a space after "1" will have correct syntax highlighting. However, without the space character, the syntax highlighting for everything that follows will be incorrect.
Screen Shot 2022-09-15 at 4 18 06 PM
Screen Shot 2022-09-15 at 4 18 13 PM

VSCode extension version: v2.5.0
Dafny version: v3.8.0

MikaelMayer:
Change the highlight grammar and added a test to fix #269

Now:
image

### DIFF
--- a/src/test/highlight/regression-highlight.dfy
+++ b/src/test/highlight/regression-highlight.dfy
@@ -1,3 +1,12 @@
+// Issue #269
+method SpaceAFterOperator(n: nat)
+  requires 1<n
+  // this should be a comment and ensures should be the same color as requires
+  ensures true
+{
+  assert 1<(2+2);
+}
+
 // Issue #240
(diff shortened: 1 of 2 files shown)


prvshah51 (approved):
I checked with test file and it's working.

developStorm:
Appriciate the fix! When can we expect to see a new release includes this?

MikaelMayer:
Do you have something urgent that requires this fix?

PR #284, opened 2022-10-12
Fix: Make brew update optional

MikaelMayer:
We figured out that `brew update` could be a bottleneck for some users so this PR makes it optional


atomb (approved):
Looks great!

prvshah51 (approved):

PR #285, opened 2022-10-13
Release v2.8.5

MikaelMayer:
Making `brew update` optional is critical for some of our users, so I would not delay publishing this update.


prvshah51 (approved):
Looks good.

PR #286, opened 2022-10-14
Only require Java to be installed when using a Dafny version before 3.9

keyboardDrummer:
### Changes

- Only build the language server when building Dafny from source for the IDE
- The Dafny language server 3.9 does not require Java to be build, so the Java check can be removed for version 3.9 and up

### Testing

Ran this on a M1 mac where previously Java could not be found from the VSCode Dafny plugin, and now it built from from source correctly.

### DIFF
--- a/src/language/dafnyInstallation.ts
+++ b/src/language/dafnyInstallation.ts
@@ -235,7 +235,7 @@ export class DafnyInstaller {
     processChdir(Utils.joinPath(installationPath, 'dafny').fsPath);
     await this.execLog('git fetch --all --tags');
     await this.execLog(`git checkout v${configuredVersion}`);
-    await this.execLog('make exe');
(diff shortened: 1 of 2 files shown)


src/language/dafnyInstallation.ts
```diff
@@ -212,25 +213,29 @@ export class DafnyInstaller {
...
+      try {
+
+        const process = await this.execLog('javac -version');
+        if(!(/javac \d+\.\d+/.exec(process.stdout))
+          && !(/javac \d+\.\d+/.exec(process.stderr))) {
+          throw '';
+        }
+      } catch(error: unknown) {
+        const errorMsg = error === '' ? 'Javac not found' : `${error}`;
+        this.writeStatus(`${errorMsg}. Please install a valid JDK`
```
MikaelMayer, line 227:
```suggestion
        this.writeStatus(`${errorMsg}. Javac is needed because you use a version of Dafny older than 3.9.0. Please install a valid JDK`
```

PR #287, opened 2022-10-17
fix: `dafny.dotnetExecutablePath` works to install custom source

seebees:
If `dotnet` is installed
and placed in `dafny.dotnetExecutablePath`
custom source should still be able to be built.

To do this `checkSupportedDotnetVersion`
needs to be the primary way to detect `dotnet`.
This has the side effect of both not double installing from `brew` but also not requireing an install if `dotnet` is already installed.

The `wget` call is also removed
since there are already helpers to perform this get and unzip. They are updated to take a `thing` to facilitate logging. There is also a bit of update to the `Z3` section
to account for this change.

The largest issue with this was
`Source/DafnyCore/DafnyCore.csproj`
has direct `dotnet` build targets.
This means that call to build will fail
if `dotnet` is not on the `PATH`.
This is unfortunate, but…
So we add it for that specific command.

Finnaly, speed up the `git clone`
to both checkout the tag and not get history.

Added a few things to `.gitignore`
and simplified the ReadMe’s use of `vsce`

### DIFF
--- a/src/language/dafnyInstallation.ts
+++ b/src/language/dafnyInstallation.ts
@@ -16,7 +16,7 @@ import { exec } from 'child_process';
 import { chdir as processChdir, cwd as processCwd } from 'process';
 import fetch from 'cross-fetch';
 
-import {checkSupportedDotnetVersion, getDotnetExecutablePath} from '../dotnet'
+import { checkSupportedDotnetVersion, getDotnetExecutablePath } from '../dotnet';
 
 const execAsync = promisify(exec);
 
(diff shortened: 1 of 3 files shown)


src/language/dafnyInstallation.ts
```diff
@@ -231,24 +238,35 @@ export class DafnyInstaller {
...
-    await this.execLog(`git checkout v${configuredVersion}`);
-    await this.execLog('dotnet build Source/DafnyLanguageServer/DafnyLanguageServer.csproj');
+
+    const { path: dotnet } = await getDotnetExecutablePath();
+    // The DafnyCore.csproj has a few targets that call `dotnet` directly.
+    // If dotnet is configured in dafny.dotnetExecutablePath
+    // it MAY NOT be on the path.
+    // This will cause the build to fail.
+    // This works around this edge case.
+    const injectPath = `PATH=${path.dirname(dotnet)}:$PATH`;
```
MikaelMayer, line 252: On Windows, the path separator is ";" not ":".... but since this code is not supposed to execute on Windows, I think it's ok.

src/language/dafnyInstallation.ts
```diff
@@ -231,24 +238,35 @@ export class DafnyInstaller {
...
+    // Build the DafnyLanguageServer
+    await this.execLog(`${injectPath} ${ (await getDotnetExecutablePath()).path } build Source/DafnyLanguageServer/DafnyLanguageServer.csproj`);
```
MikaelMayer, line 254: Oh I see, that way, you can specify both the dotnet path in the settings.

MikaelMayer (approved):
Looks good ! This will ease the installation of Dafny VSCode for Mac M1 users

Issue #52, opened 2021-10-01
New visualization for counterexamples


camrein:
With the merge of https://github.com/dafny-lang/dafny/pull/1336, counterexamples may now display values of reference types. This makes inline counterexamples hard to read.
![image](https://user-images.githubusercontent.com/18049310/135620734-cc96ea1b-651b-43aa-b226-097a5426b410.png)

As suggested by @Dargones, something like a debugger view might be suitable here. As a first PoC (see [branch](https://github.com/dafny-lang/ide-vscode/tree/feature/counterexample-treeview)), I've created a tree-view that displays the variables and their values for the active line:
![image](https://user-images.githubusercontent.com/18049310/135621060-5b39d651-0f46-4bc2-8d40-766ae51d5f00.png)
Further improvements (e.g. nesting) will probably require changes to the language server.

Dargones:
This looks great, thank you for doing this! I can make the necessary changes to the language server to make nesting display possible. What format would be the most intuitive? My understanding is that one would have to edit `CounterExampleItem` class to achieve this but I am wondering what is the correct way to make these changes to preserve compatibility across versions of the plugin\server. There is also the issue of `CounterExampleProvider` in the older `DafnyServer` code - I am not certain if it is still being used and\or should be supported.

Dargones:
This looks great, thank you for doing this! I can make the necessary changes to the language server to make nesting display possible. What format would be the most intuitive? My understanding is that one would have to edit `CounterExampleItem` class to achieve this but I am wondering what is the correct way to make these changes to preserve compatibility across versions of the plugin\server. There is also the issue of `CounterExampleProvider` in the older `DafnyServer` code - I am not certain if it is still being used and\or should be supported.

camrein:
> I can make the necessary changes to the language server to make nesting display possible. What format would be the most intuitive?

I guess some tree structure makes sense since it's a TreeView (basically the same as the file explorer of VSCode).

> My understanding is that one would have to edit `CounterExampleItem` class to achieve this but I am wondering what is the correct way to make these changes to preserve compatibility across versions of the plugin\server.

[...]

PR #290, opened 2022-10-27
Allow copying counterexamples to the clipboard

dijkstracula:
This patch adds a new contextual menu option to allow copying the counterexamples received from the language server to the system clipboard.  At present, non-trivial counterexamples are hard to read in the current "everything appears in one line" form.  Additionally, they are difficult to share with colleagues or instructors, inhibiting "rubber-duck" counterexample debugging.  For instance, I have a counterexample that stretches, I'd wager, _twenty to thirty_ screen widths across in a project I'm working on now - honestly I have no idea how to parse it at that point.

I think all parties agree that #52 would go a long way to addressing the former and, depending on how VS code lets users extract state from the debugger, potentially the latter too.  However, this issue hasn't seen any activity since last year.

This patch simply introduces a new contextual menu option to copy counterexamples to the clipboard. 

image

Currently they get copied in the same order that the language server hands them to us (which is not in line order, but, perhaps it should be?  In the presence of multiple errors, though, seems like one could have collisions, so I'm leaving it like this for now)

Examples of clipboard output might be:

```
At "ensures A.Next(IC(c), I(c, s), I(c, s')) {" (file:///Users/ntaylor/school/smrt/src/BlockDisk.dfy:277):
    c:BlockDisk.Constants = Constants(n__blocks := 4294966632)
    s:BlockDisk.State = State(blocks := @1)
    s':BlockDisk.State = State(blocks := @2)
    block_id:int = 4194663
    val:seq<NativeTypes.uint8> = (Length := 1024)
    aconst:BlockDisk.Constants = ()
    astate:BlockDisk.State = ()
    astate':BlockDisk.State = ()
    byte_off:int = 0
    @1:seq<seq<NativeTypes.uint8>> = (Length := 4294966632, [4194663] := val)
    @2:seq<seq<NativeTypes.uint8>> = (Length := 4294966632)

At "var aconst := IC(c);" (file:///Users/ntaylor/school/smrt/src/BlockDisk.dfy:278):
    c:BlockDisk.Constants = Constants(n__blocks := 4294966632, n__blocks := 4294966632)
    s:BlockDisk.State = State(blocks := @1)
    s':BlockDisk.State = State(blocks := @2)
    block_id:int = 4194663
    val:seq<NativeTypes.uint8> = (Length := 1024)
    aconst:BlockDisk.Constants = c
    astate:BlockDisk.State = ()
    astate':BlockDisk.State = ()
    byte_off:int = 0
    @1:seq<seq<NativeTypes.uint8>> = (Length := 4294966632, [4194663] := val)
    @2:seq<seq<NativeTypes.uint8>> = (Length := 4294966632)
```

It's pretty undecorated - I wanted to leave it essentially in the same form as the extension shows, but with just enough structure to make it more readable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -268,7 +268,7 @@
       },
       {
         "command": "dafny.copyCounterexamples",
-        "title": "Dafny: Copy All Counterexamples to Clipboard"
+        "title": "Dafny: Copy Counterexamples to Clipboard (Experimental)"
       },
       {
         "command": "dafny.restartServer",
--- a/package.json
+++ b/package.json
@@ -52,6 +52,11 @@
           "when": "editorTextFocus && editorLangId == dafny && !dafny.counterexampleMenu.CanShowCounterExample",
           "command": "dafny.hideCounterexample",
           "group": "7_Dafny@2"
+        },
+        {
+          "when": "editorTextFocus && editorLangId == dafny",
+          "command": "dafny.copyCounterexamples",
+          "group": "7_Dafny@2"
         }
       ]
     },
@@ -260,6 +265,10 @@
       {
         "command": "dafny.hideCounterexample",
         "title": "Dafny: Hide Counterexample"
+      },
+      {
+        "command": "dafny.copyCounterexamples",
+        "title": "Dafny: Copy All Counterexamples to Clipboard"
       },
       {
         "command": "dafny.restartServer",
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -4,6 +4,7 @@ export namespace DafnyCommands {
   export const CompileAndRun = 'dafny.compileAndRun';
   export const ShowCounterexample = 'dafny.showCounterexample';
   export const HideCounterexample = 'dafny.hideCounterexample';
+  export const CopyCounterexamples = 'dafny.copyCounterexamples';
   export const OpenStatusBarMenu = 'dafny.openStatusBarMenu';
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -261,6 +266,10 @@
         "command": "dafny.hideCounterexample",
         "title": "Dafny: Hide Counterexample"
       },
+      {
+        "command": "dafny.copyCounterexamples",
+        "title": "Dafny: Copy All Counterexamples to Clipboard"
```
MikaelMayer, line 271:
```suggestion
        "title": "Dafny: Copy Counterexamples to Clipboard (Experimental)"
```

MikaelMayer (changes requested):
This is definitely more useful than what we have right now.
A minor point though we should warn our users that this is experimental, as it's very likely that the next iteration will not copy things to the clipboard, but will display them instead in a better view.

package.json
dijkstracula: That's a good point, thanks!

MikaelMayer (approved):
I'm looking forward to a better UI, but for now, that's strictly better.
I would love to get the approval of @keyboardDrummer  as well.

dijkstracula:
Makes sense; I won't feel offended when you rip it out and replace it with the better thing :-)

keyboardDrummer (approved):

PR #291, opened 2022-10-27
Release v2.8.6 (updated Dafny to 3.9.1)

atomb:


prvshah51:
Need one more change , constant `LatestVersion`  in `src/constant.ts` has to be updated with latest Dafny version number.

rod-chapman:
Does
 export const LatestVersion = '3.9.0';
also need to change in src/constants.ts
??

rod-chapman:
You should probably add comments in package.json and src/constants.ts to remind people that if one changes, then so does the other...

MikaelMayer:
> You should probably add comments in package.json and src/constants.ts to remind people that if one changes, then so does the other...

There is nothing to remind about, `./publish-process.sh` should take care of everything. Let me see what went wrong.
https://github.com/dafny-lang/ide-vscode/blob/108c6ad8ca49c4fb0bba9db3516ae4f43165d6cd/publish_process.js#L224-L229

MikaelMayer (changes requested):
There are a few more places to change. I'll reach out to you directly.

atomb:
Ah, yeah, this was my experiment at what blindly following the release documentation for Dafny would produce. :) It looks like `release_process.ts` needs a few tweaks. But I can update this PR to include those changes manually, for this release.

MikaelMayer (approved):

PR #292, opened 2022-10-28
Fix the Dafny updating mechanism in publish_process.js

MikaelMayer:
This PR fixes publish_process.js by ensuring constant.ts is correctly updated. I tested this PR locally and now it works.


atomb (approved):
Thanks for fixing this so quickly!

PR #297, opened 2022-11-02
Fix: No longer direct dependency to DafnyLanguageServer.dll

MikaelMayer:
Previously, when setting a custom path to DafnyLanguageServer.dll, it was not possible to overwrite the dll while VSCode was active. This PR makes it possible to do so.

### DIFF
--- a/src/language/dafnyInstallation.ts
+++ b/src/language/dafnyInstallation.ts
@@ -89,15 +89,41 @@ export async function getCompilerRuntimePath(context: ExtensionContext): Promise
   return configuredPath;
 }
 
+// We cache the language server runtime path so that we don't need to copy it every time.
+let LanguageServerRuntimePath: string | null = null;
+
 export async function getLanguageServerRuntimePath(context: ExtensionContext): Promise<string> {
-  const configuredPath = await ifNullOrEmpty(
-    getConfiguredLanguageServerRuntimePath(),
+  if(LanguageServerRuntimePath != null) {
+    return LanguageServerRuntimePath;
+  }
+  const configuredLanguageServerPath = getConfiguredLanguageServerRuntimePath();
+  let mightBeRecomputed = false;
+  if(configuredLanguageServerPath !== '' && configuredLanguageServerPath != null) {
+    mightBeRecomputed = true;
+  }
+
+  let configuredPath = await ifNullOrEmpty(
+    configuredLanguageServerPath,
     async () => LanguageServerConstants.GetDefaultPath(await getConfiguredVersion(context))
   );
-  if(path.isAbsolute(configuredPath)) {
-    return configuredPath;
+  if(!path.isAbsolute(configuredPath)) {
+    configuredPath = path.join(context.extensionPath, configuredPath);
   }
-  return path.join(context.extensionPath, configuredPath);
+  if(mightBeRecomputed && (/\.dll$/.exec(configuredPath)) != null) {
+    // We copy DafnyLanguageServer.dll to another location
+    // so that rebuilding Dafny will not fail because it's open in VSCode
+    const newPath = configuredPath.replace('.dll', '-vscode.dll');
+    await fs.promises.copyFile(configuredPath, newPath);
+    const oldDepsJson = configuredPath.replace('.dll', '.deps.json');
+    const newDepsJson = configuredPath.replace('.dll', '-vscode.deps.json');
+    await fs.promises.copyFile(oldDepsJson, newDepsJson);
+    const oldRunJson = configuredPath.replace('.dll', '.runtimeconfig.json');
+    const newRunJson = configuredPath.replace('.dll', '-vscode.runtimeconfig.json');
+    await fs.promises.copyFile(oldRunJson, newRunJson);
+    configuredPath = newPath;
+  }
+  LanguageServerRuntimePath = configuredPath;
+  return configuredPath;
 }
 
 function getConfiguredLanguageServerRuntimePath(): string {


MikaelMayer:
Well, looks like this PR is not complete because DafnyCore.dll still cannot be overwritten

src/language/dafnyInstallation.ts
```diff
@@ -89,15 +89,41 @@ export async function getCompilerRuntimePath(context: ExtensionContext): Promise
...
-  const configuredPath = await ifNullOrEmpty(
-    getConfiguredLanguageServerRuntimePath(),
+  if(LanguageServerRuntimePath != null) {
+    return LanguageServerRuntimePath;
+  }
+  const configuredLanguageServerPath = getConfiguredLanguageServerRuntimePath();
+  let mightBeRecomputed = false;
+  if(configuredLanguageServerPath !== '' && configuredLanguageServerPath != null) {
+    mightBeRecomputed = true;
+  }
```
camrein, line 103: You may want to extract the empty string check into a separate function. IIRC, we're doing this in multiple occasions so there might be an opportunity to introduce/re-use a common function.

Moreover, you could simplify the expression like this:
```suggestion
  let mightBeRecomputed = configuredLanguageServerPath !== '' && configuredLanguageServerPath != null;
```

src/language/dafnyInstallation.ts
```diff
@@ -89,15 +89,41 @@ export async function getCompilerRuntimePath(context: ExtensionContext): Promise
...
+    configuredLanguageServerPath,
     async () => LanguageServerConstants.GetDefaultPath(await getConfiguredVersion(context))
   );
-  if(path.isAbsolute(configuredPath)) {
-    return configuredPath;
+  if(!path.isAbsolute(configuredPath)) {
+    configuredPath = path.join(context.extensionPath, configuredPath);
   }
-  return path.join(context.extensionPath, configuredPath);
+  if(mightBeRecomputed && (/\.dll$/.exec(configuredPath)) != null) {
```
camrein, line 112: I suggest extracting the checking part into a function to make the intent clear to the reader.

For example:
```typescript
function isDll(path: string) {
  return /\.dll$/.test(configuredPath);
}
```

As you see, using `test(...)` simplifies the call since you don't have to check against `null`. That's possible since you only want to check if the pattern matches, not extracting the matching bits.

Last but not least, you may want to replace the regex entirely. Currently, you only want to check the suffix of the string, so you could use `endsWith` instead. In my opinion, that's more readable than a regex as long as it's a simple test like this:
For example:
```typescript
function isDll(path: string) {
  return configuredPath.endsWith('.dll');
}
```


PS: You may want to account for different cases. Then you may use your regex with `/i` to ignore the casing of the string.

src/language/dafnyInstallation.ts
```diff
@@ -89,15 +89,41 @@ export async function getCompilerRuntimePath(context: ExtensionContext): Promise
...
+    // We copy DafnyLanguageServer.dll to another location
+    // so that rebuilding Dafny will not fail because it's open in VSCode
+    const newPath = configuredPath.replace('.dll', '-vscode.dll');
```
camrein, line 115: Careful, this will replace only the first occurence of `.dll` in the string, not the last. So if there are multiple `.dll` bits within the string, you'll end up with something different.
In this case, you can go back to your regex approach to make sure it only replaces the endling `.dll`, Something like this:

```suggestion
    const newPath = configuredPath.replace(/\.dll$/, '-vscode.dll');
```

src/language/dafnyInstallation.ts
```diff
@@ -89,15 +89,41 @@ export async function getCompilerRuntimePath(context: ExtensionContext): Promise
...
+    await fs.promises.copyFile(configuredPath, newPath);
+    const oldDepsJson = configuredPath.replace('.dll', '.deps.json');
+    const newDepsJson = configuredPath.replace('.dll', '-vscode.deps.json');
+    await fs.promises.copyFile(oldDepsJson, newDepsJson);
+    const oldRunJson = configuredPath.replace('.dll', '.runtimeconfig.json');
+    const newRunJson = configuredPath.replace('.dll', '-vscode.runtimeconfig.json');
+    await fs.promises.copyFile(oldRunJson, newRunJson);
+    configuredPath = newPath;
```
camrein, line 123: I suggest extracting this whole duplication logic into a separate function.

camrein (changes requested):
This is a clever workaround to address this particular problem! I left some comments to improve your changes.

src/language/dafnyInstallation.ts
```diff
@@ -24,7 +24,7 @@ const ArchiveFileName = 'dafny.zip';
 const mkdirAsync = promisify(fs.mkdir);
 
 // Equivalent to a || b but without ESLint warnings
-async function ifNullOrEmpty(a: string | null, b: () => Promise<string>): Promise<string> {
+async function ifNullOrEmptyAsync(a: string | null, b: () => Promise<string>): Promise<string> {
```
camrein, line 27: nit: In contrast to C#, the `Async` suffix is rather uncommon in JavaScript/TypeScript. Would make a few things better though. :)

src/language/dafnyInstallation.ts
```diff
@@ -89,15 +89,63 @@ export async function getCompilerRuntimePath(context: ExtensionContext): Promise
...
+}
+
 export async function getLanguageServerRuntimePath(context: ExtensionContext): Promise<string> {
-  const configuredPath = await ifNullOrEmpty(
-    getConfiguredLanguageServerRuntimePath(),
+  if(LanguageServerRuntimePath != null) {
+    return LanguageServerRuntimePath;
+  }
+  const configuredLanguageServerPath = getConfiguredLanguageServerRuntimePath();
+  const mightBeRecomputed: boolean = !isNullOrEmpty(configuredLanguageServerPath);
```
camrein, line 104: nit: The type-specification is unnecessary here.

```suggestion
  const mightBeRecomputed = !isNullOrEmpty(configuredLanguageServerPath);
```

src/language/dafnyInstallation.ts
```diff
@@ -89,15 +89,63 @@ export async function getCompilerRuntimePath(context: ExtensionContext): Promise
...
-  return path.join(context.extensionPath, configuredPath);
+  LanguageServerRuntimePath = configuredPath;
+  return configuredPath;
+}
+
+async function cloneAllNecessaryDlls(configuredPath: string): Promise<string> {
+  const DLSName = 'DafnyLanguageServer';
+  const DCName = 'DafnyCore';
+  const dll = '.dll';
+  const DLS = DLSName + dll;
```
camrein, line 124: nit: Keep the style-convention for local variables and start with lower-case letters. For example:
```suggestion
  const dlsName = 'DafnyLanguageServer';
  const dcName = 'DafnyCore';
  const dll = '.dll';
  const dls = dlsName + dll;
```

src/language/dafnyInstallation.ts
```diff
@@ -89,15 +89,63 @@ export async function getCompilerRuntimePath(context: ExtensionContext): Promise
...
+  if(configuredPath.endsWith(DLS)) {
+    const prefix = configuredPath.substring(0, configuredPath.length - DLS.length);
```
camrein, line 126: nit: This might be more suitable here:
```suggestion
    const prefix = path.dirname(configuredPath);
```

And then using it like this:
```
const newPath = path.join(prefix);
```

Great opportunity to rename `prefix` to something like `installationDir` too.

camrein (approved):
Great!

src/language/dafnyInstallation.ts
```diff
@@ -89,15 +89,63 @@ export async function getCompilerRuntimePath(context: ExtensionContext): Promise
...
   }
-  return path.join(context.extensionPath, configuredPath);
+  if(mightBeRecomputed) {
+    configuredPath = await cloneAllNecessaryDlls(configuredLanguageServerPath);
+  }
+  LanguageServerRuntimePath = configuredPath;
+  return configuredPath;
+}
+
+async function cloneAllNecessaryDlls(configuredPath: string): Promise<string> {
```
camrein, line 120: neat!

src/language/dafnyInstallation.ts
```diff
@@ -89,15 +89,63 @@ export async function getCompilerRuntimePath(context: ExtensionContext): Promise
...
+    const oldDepsJson = path.join(installationDir, `${dlsName}.deps.json`);
+    const newDepsJson = path.join(installationDir, `${dlsName}-vscode.deps.json`);
+    const oldContent = await fs.promises.readFile(oldDepsJson, { encoding: 'utf8' });
+    const newContent = oldContent.replace(/DafnyCore\.dll/g, 'vscode-dll/DafnyCore.dll');
+    await fs.promises.writeFile(newDepsJson, newContent);
+    const oldRunJson = path.join(installationDir, `${dlsName}.runtimeconfig.json`);
+    const newRunJson = path.join(installationDir, `${dlsName}-vscode.runtimeconfig.json`);
+    await fs.promises.copyFile(oldRunJson, newRunJson);
+    const vscodedllpath = path.join(installationDir, 'vscode.dll');
+    if(!fs.existsSync(vscodedllpath)) {
```
camrein, line 140: Avoid the use of blocking sync methods within javascript code. They can lead to freezing UIs, making it a bad user experience.

src/language/dafnyInstallation.ts
```diff
@@ -89,15 +89,63 @@ export async function getCompilerRuntimePath(context: ExtensionContext): Promise
...
 }
 
+// We cache the language server runtime path so that we don't need to copy it every time.
+let LanguageServerRuntimePath: string | null = null;
+
+function isNullOrEmpty(s: string | null): boolean {
+  return s === '' || s == null;
+}
+
 export async function getLanguageServerRuntimePath(context: ExtensionContext): Promise<string> {
```
camrein, line 99: I want to point out here that you accidentally introduced a race condition here due to the asynchronous implementation.
It's possible that `LanguageServerRuntimePath` changes at any `await`.

Moreover: I suggest separating the `getLanguageServerRuntimePath` functionallity from the copying functionality. This function has 2 primary uses: Checking/retrieving the current Dafny installation when starting up as well as identifying the target directory for the automatic installation. The latter would fail but it looks like you've already respected this case with `mightBeRecomputed`. Maybe naming this variable to something like `isCustomInstallation` is more informative here.
Personally, I don't like `get*` functions that have such side-effects as most people wouldn't expect it, especially due to the name. I'm sure you can come up with a good idea to separate it since the only location where this is of any concern is in `dafnyLanguageClient.ts`.

camrein (approved):

src/language/dafnyInstallation.ts
MikaelMayer: > It's possible that LanguageServerRuntimePath changes at any await.

No it's not in this case, because JS is single-threaded, and all the calls to getLanguageServerRuntimePath are within the same function call.

I renamed `getLanguageServerRuntimePath` to `getOrComputeLanguageServerRuntimePath` to indicate potential side-effects, and I would advise against anything that looks like we have to explicitly manage at which point we do the duplication of the dlls, and at which point we only retrieve the computed value. It would force adding parameters where we don't want to, or come up with class fields that can be nullable and then we have to deal with various warnings.
`getLanguageServerRuntimePath` is encapsulating the behavior, and it's not modifying anything a user of this function can access, so it's still a true get method.

camrein (approved):

PR #298, opened 2022-11-03
Follow-up: Copy everything, not just one dll

MikaelMayer:


src/language/dafnyInstallation.ts
```diff
@@ -120,30 +136,41 @@ export async function getOrComputeLanguageServerRuntimePath(context: ExtensionCo
...
+    // Copy all the files from installationDir to vscodeDir
+    const files = await fs.promises.readdir(installationDir);
+    for(const file of files) {
+      if(!(file.endsWith(dll)
+      || file.endsWith(runtimeconfigjson)
+      || file.endsWith(depsjson)
+      || file.endsWith('.pdb')
+      || file === 'z3')) {
+        continue;
+      }
```
camrein, line 160: nit: I suggest extracting a method here or at least indent the different cases.

src/language/dafnyInstallation.ts
```diff
@@ -120,30 +136,41 @@ export async function getOrComputeLanguageServerRuntimePath(context: ExtensionCo
...
+      // If it's a directory, we use the function above
+      if((await fs.promises.stat(path.join(installationDir, file)))
+        .isDirectory()) {
```
camrein, line 163: nit: In my opinion, it looks cleaner if the whole condition is on a single line.
```suggestion
      if((await fs.promises.stat(path.join(installationDir, file))).isDirectory()) {
```

src/language/dafnyInstallation.ts
```diff
@@ -120,30 +136,41 @@ export async function getOrComputeLanguageServerRuntimePath(context: ExtensionCo
...
+        await copyDir(path.join(installationDir, file),
+          path.join(vscodeDir, file));
```
camrein, line 165: nit
```suggestion
        await copyDir(path.join(installationDir, file), path.join(vscodeDir, file));
```

camrein: Alternatively, if you prefer multiline, I like to place all arguments on a separate line:
```suggestion
        await copyDir(
          path.join(installationDir, file),
          path.join(vscodeDir, file)
        );
```
Similar to having a nested block of code.

src/language/dafnyInstallation.ts
```diff
@@ -120,30 +136,41 @@ export async function getOrComputeLanguageServerRuntimePath(context: ExtensionCo
...
-    const newRunJson = path.join(installationDir, `${dlsName}-vscode.runtimeconfig.json`);
-    await fs.promises.copyFile(oldRunJson, newRunJson);
-    const vscodedllpath = path.join(installationDir, 'vscode.dll');
-    if(!await existsAsync(vscodedllpath)) {
-      await fs.promises.mkdir(vscodedllpath);
-    }
-    const oldCore = path.join(installationDir, `${dcName}${dll}`);
-    const newCore = path.join(vscodedllpath, `${dcName}${dll}`);
-    await fs.promises.copyFile(oldCore, newCore);
+    const newPath = path.join(vscodeDir, `${dlsName}.dll`);
```
camrein, line 173: nit: I have the feeling that his comment is relevant for the whole copy procedure and not only this particular dll.

src/language/dafnyInstallation.ts
```diff
@@ -97,6 +97,22 @@ function isNullOrEmpty(s: string | null): boolean {
   return s === '' || s == null;
 }
 
+async function copyDir(src: string, dest: string) {
```
camrein, line 100: nice!

MikaelMayer:
I'm puzzled by a new problem. If I have two VSCode windows open, it will now try to copy the dlls as well, but it won't be able because the dlls are locked by the other version.

PR #301, opened 2022-11-07
Ghost code more like comments

MikaelMayer:
We have received the feedback that highlights of ghost code can sometimes be confused for a kind of selection. This PR solves the problem by changing the style of ghost code to be
- Italics: Reminds of comments.
- Opacity: Makes it immediately appearing where the ghost code is or if some ambiguous code is ghost.

## Before:

|        | Dark | Light |
| -----|-------------------|---------------|
| Old, without selection | image | image |
| Old, with selection | image | image |
|| **This PR would do the following** ||
| New, without selection | image | image |
[ New, with selection | image | image |

These two visual cues would not interfere with text selection at all, are reminiscent of other specifications like in C#, and feel more natural to me.

### DIFF
--- a/src/ui/ghostDiagnosticsView.ts
+++ b/src/ui/ghostDiagnosticsView.ts
@@ -6,11 +6,12 @@ import { DafnyLanguageClient } from '../language/dafnyLanguageClient';
 import { getVsDocumentPath, toVsRange } from '../tools/vscode';
 
 const GhostDecoration: DecorationRenderOptions = {
+  fontStyle: 'italic',
   dark: {
-    backgroundColor: '#64646480'
+    opacity: '80%'
   },
   light: {
-    backgroundColor: '#D3D3D380'
+    opacity: '70%'
   }
 };
 


cpitclaudel:
Can you show how this looks for a large block of spec code?  E.g. a `while` loop with many invariants or a large `calc` or `forall` block

MikaelMayer:
Where did my post go? I had written a big post with examples. Let me try again
image

MikaelMayer:
It's also pretty neat in functions

image

cpitclaudel:
Works for me!  But I'm not sure the italics are necessary (the opacity change is already pretty good?)
By the way, why is `ghost var j` not highlighted?

alex-chew:
IMHO the opacity difference by itself doesn't provide enough contrast; without the additional italic styling, it's hard for me to tell the ghost code apart from the non-ghost code unless I really try to look for it.

MikaelMayer:
> Works for me! But I'm not sure the italics are necessary (the opacity change is already pretty good?)
> By the way, why is `ghost var j` not highlighted?

Because it's not ambiguous that it's ghost... Yes I know, inconsistent, and I wish we had more options about that. But that's a different story

MikaelMayer:
> IMHO the opacity difference by itself doesn't provide enough contrast; without the additional italic styling, it's hard for me to tell the ghost code apart from the non-ghost code unless I really try to look for it.

Exactly.

keyboardDrummer (approved):

PR #309, opened 2022-11-17
Fix: Building from source also builds Dafny.dll

MikaelMayer:
This PR ensures that, on Mac installations that cannot use the released binaries, it builds also Dafny.dll, not just DafnyLanguageServer.dll


keyboardDrummer:
This fixes a bug but also introduces a Java requirement which sometimes prevents the whole installation from completing. It might be easier to wait for the PR that lets VSCode use the published arm64 binary.

keyboardDrummer (approved):
Could you describe what testing you did ?

src/language/fromSourceInstaller.ts
```diff
@@ -104,9 +104,20 @@ export class FromSourceInstaller {
...
-    const projectToBuild = configuredVersionToNumeric(configuredVersion) < configuredVersionToNumeric('3.10.0')
-      ? 'DafnyLanguageServer/DafnyLanguageServer.csproj'
-      : 'DafnyDriver/DafnyDriver.csproj';
+    // If java is not installed, remove from Source/DafnyRuntime/DafnyRuntime.csproj
+    // the section <Target Name="BuildDafnyRuntimeJar">...</Target>
+    try {
+      await this.execLog(`${injectPath} java -version`);
+    } catch(e: unknown) {
+      // Read the project file, and then remove the section
+      let DafnyRunTimeCsprojContent = await fs.promises.readFile(Utils.joinPath(installationPath, 'dafny', 'Source', 'DafnyRuntime', 'DafnyRuntime.csproj').fsPath, 'utf8');
```
keyboardDrummer, line 113: Maybe this whole block should be in a try/catch as well, so that if it fails we still try to build.

src/language/fromSourceInstaller.ts
```diff
@@ -104,9 +104,20 @@ export class FromSourceInstaller {
...
+      DafnyRunTimeCsprojContent = DafnyRunTimeCsprojContent.replace(/<Target Name="BuildDafnyRuntimeJar">[\s\S]*?<\/Target>/, '');
```
MikaelMayer, line 114:
```suggestion
      DafnyRunTimeCsprojContent = DafnyRunTimeCsprojContent.replace(/<Target Name="BuildDafnyRuntimeJar"[\s\S]*?<\/Target>/, '');
```

MikaelMayer:
> Cool change. Thanks!
> 
> Could you describe what testing you did ?

Lightweight tests:
1. I tested that dotnet build still works on the Dafny runtime and does not trigger the build of the jar after removing the two elements
2. I tested the regexes in regex101 to ensure they match (and I even found a typo, so thanks for the comment)

src/language/fromSourceInstaller.ts
MikaelMayer: I suggest we simply update the script as needed when the time will come. Since it's a fallback script, we don't have a clear case anyway

PR #310, opened 2022-11-18
Support two instances of VSCode without crashing.

MikaelMayer:
This PR ensures that two instances of VSCode can run Dafny on a locally built DafnyLanguageServer.dll without the second crashing like this
image

VSCode will basically detect a custom installation, and copy all the necessary dependencies to a unique temporary folder. This enables that:
- Developers writing Dafny code can also continuously compile the DLLs without having an error that the DLL are being used, e.g. they can also run tests in their IDE.
- There can be multiple instances of VSCode running.
- Reloading VSCode always copies the files again, so one can simply reload VSCode after compiling the language server.

### DIFF
--- a/src/language/dafnyInstallation.ts
+++ b/src/language/dafnyInstallation.ts
@@ -98,37 +98,27 @@ function isNullOrEmpty(s: string | null): boolean {
 }
 
 async function ensureDirExists(
-  dirName: string,
-  addCleanup: (callback: () => Promise<void>) => void,
-  forceDeletingOnCleanupEvenIfExisted: boolean = false) {
+  dirName: string) {
   const existed = await existsAsync(dirName);
   if(!existed) {
     await mkdirAsync(dirName);
   }
-  if(!existed || forceDeletingOnCleanupEvenIfExisted) {
-    addCleanup(async () => {
-      await fs.promises.rmdir(dirName);
-    });
-  }
 }
 
-async function copyFile(srcPath: string, targetFile: string, addCleanup: (callback: () => Promise<void>) => void) {
+async function copyFile(srcPath: string, targetFile: string) {
   await fs.promises.copyFile(srcPath, targetFile);
-  addCleanup(((targetFile: string) => async () => {
-    await fs.promises.rm(targetFile);
-  })(targetFile));
 }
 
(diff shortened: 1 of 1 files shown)


src/language/dafnyInstallation.ts
```diff
@@ -142,40 +162,49 @@ async function cloneAllNecessaryDlls(configuredPath: string): Promise<string> {
...
-    const vscodeDir = path.join(installationDir, 'vscode');
-    // If the folder does not exist, create it
-    if(!await existsAsync(vscodeDir)) {
-      await mkdirAsync(vscodeDir);
+  if(!configuredPath.endsWith(dls)) {
+    return configuredPath;
+  }
+  const installationDir = path.dirname(configuredPath);
+  // copy all the files from installationDir to a new temporary folder
+  const vscodeDir = path.join(os.tmpdir(), await fs.promises.mkdtemp('vscode-dafny-dlls-'));
```
robin-aws, line 170: Could you just register one callback to delete this directory and everything in it here (using `recursive = true` in the call to `rmDir`), rather than every intermediate directory separately? Then you don't need the complexity of threading that through `ensureDirExists`.

robin-aws (commented):
Great, the temporary directory idea seems to work really nicely. Besides the simplification suggestion, make sure you update the PR description since the behavior has changed (for the better!)

src/language/dafnyInstallation.ts
```diff
@@ -142,40 +152,44 @@ async function cloneAllNecessaryDlls(configuredPath: string): Promise<string> {
...
+      await copyFile(srcFile, targetFile);
     }
-
-    const newPath = path.join(vscodeDir, `${dlsName}.dll`);
-    return newPath;
   }
-  return configuredPath;
+
+  const newPath = path.join(vscodeDir, dls);
+  process.on('exit', cleanup);
```
robin-aws, line 191: For the record, it would be slightly better to register the cleanup immediately after creating the temporary directory, just to make the window where an early return/crash would defeat the mechanism smaller.

But given this is just a best effort on top of the OS temporary file clean up, I won't block on it. :)

robin-aws (approved):
🚀

PR #312, opened 2022-12-01
Add (experimental) to the title of the counterexample commands

keyboardDrummer:


package.json
```diff
@@ -261,11 +261,11 @@
       },
       {
         "command": "dafny.showCounterexample",
-        "title": "Dafny: Show Counterexample"
+        "title": "Dafny: (experimental) Show Counterexample"
```
MikaelMayer, line 264: I think too it's a good idea to mark this feature as experimental.

I'm a bit afraid it will make the menu too large. Is it ok when you test it visually ?
If not, could we consider removing "Dafny; " from the title?

keyboardDrummer: image

MikaelMayer: I mean, when you right click on the file content? That's usually how I run counter-examples.

keyboardDrummer: image

keyboardDrummer: I'd prefer taking these out of the context menu entirely, since they don't relate to the context.

MikaelMayer: Oh no please don't remove them. This is what makes these experimental commands discoverable. A lot of users like me barely use commands and prefer click and point interfaces. I don't even know the hotkey to open the command line !

Here is what I precisely think I want that will make me comfortable to approve this PR.
- Keep the name of the commands the way you want
- Create a menu named "Dafny" and put all the 5 commands inside them.

That will have the effect to reduce the context menu to an appropriate size, while keeping the Dafny: and (experimental) keywords
image
To do so, you can modify the `package.json` this way:

```diff
  "contributes": {
+    "submenus": [
+      {
+        "id": "dafny.context",
+        "label": "Dafny"
+      }
+    ],
     "menus": {
       "editor/context": [
+        {
+          "when": "editorTextFocus && editorLangId == dafny",
+          "submenu": "dafny.context",
+          "group": "7_Dafny@8"
+        }
+      ],
+      "dafny.context": [
         {
           "when": "editorTextFocus && editorLangId == dafny",
           "command": "dafny.compile",
           "group": "7_Dafny@5"
         },
```

keyboardDrummer: Works for me, thanks.

MikaelMayer (approved):
Perfect, thanks for doing the change!

PR #314, opened 2022-12-07
Use dafny server

keyboardDrummer:
### Changes
- Use the Dafny CLI and its `server` command to start a language server
- Merge the `languageServerRuntimePath` and `compilerRuntimePath` options into one `cliPath` option.
- Move code for installing from source, copying Dafny dlls, and installing a standalone Dafny language server, into separate files

### Testing
- `$DAFNY_DEV_SERVER` configured to the latest Dafny CLI, works using `dafny server`
- Using `latest nightly` as the preferred version, where the latest nightly is of version 3.10, works using `dafny server`
- Using `latest` as the preferred version, works using the standalone Dafny server, also when it still has to be installed
- Using `3.9` as the preferred version, works using the standalone Dafny server, 
  - When it's already installed
  - When it still has to be installed
- Using `3.9` as the preferred version, works using the standalone Dafny server, also when it still has to be installed from source since the machine is an M1 Macbook.
- Test 'compile' and 'compile and run' commands, using >= 3.7 and <3.7 versions
  - Test compiler args
- Test language server launch args such as `--warn-shadowing`
  - When using dafny server
  - When using standalone server (has no effect)


src/language/dafnyInstallation.ts
```diff
@@ -1,477 +1,163 @@
...
+    case LanguageServerConstants.LatestStable: {
+      const version = LanguageServerConstants.LatestVersion;
+      toolVersion = versions.filter(l => l.startsWith(version))[0];
+      window.showInformationMessage(`Using latest stable version: ${toolVersion}`);
+      break;
+    }
+    case LanguageServerConstants.LatestNightly: {
+      const nightlies = versions.filter(l => l.includes('nightly'));
+      const dates: { index: number, date: string }[] = nightlies.map((n, index) => {
+        const split: string[] = n.split('-');
```
MikaelMayer, line 126: Could you please add a comment about what to expect here, a typical string that is being splitted?

src/language/dafnyInstallation.ts
```diff
@@ -1,477 +1,163 @@
...
 
-    // Clone the right version
-    await this.execLog(`git clone -b v${configuredVersion} --depth 1 --recurse-submodules ${LanguageServerConstants.DafnyGitUrl}`);
-    processChdir(Utils.joinPath(installationPath, 'dafny').fsPath);
+  private static async getDotnetToolVersionUncached(): Promise<string> {
+    const { path: dotnetExecutable } = await getDotnetExecutablePath();
+    const { stdout } = await execFileAsync(dotnetExecutable, [ 'tool', 'search', 'Dafny', '--detail', '--prerelease' ]);
+    const entries = stdout.split('----------------').map(entry => entry.split('\n').filter(e => e !== ''));
+    const dafnyEntry = entries.filter(entry => entry[0] === 'dafny')[0];
+    const versions = dafnyEntry.slice(8).map(versionLine => versionLine.trimStart().split(' ')[0]);
```
MikaelMayer, line 112: Rather than slice(8), could you write something like slice("astring".length)? That way, we would know what you are slicing away.

src/language/dafnyInstallation.ts
```diff
@@ -1,477 +1,163 @@
...
-        await workspace.fs.delete(archive, { useTrash: false });
-        this.writeStatus('Dafny installation completed');
-        return true;
-      }
+      await mkdir(localToolPath, { recursive: true });
+      await execFileAsync(dotnetExecutable, [ 'new', 'tool-manifest' ], { cwd: localToolPath });
+      this.statusOutput.show();
+      window.showInformationMessage(Messages.Installation.Start);
+      this.writeStatus(Messages.Installation.Start);
+      await execFileAsync(dotnetExecutable, [ 'tool', 'install', 'Dafny', '--version', toolVersion ], { cwd: localToolPath });
```
MikaelMayer, line 86: Wow it blows my mind. You just get the dotnet tool install instead of fetching binaries. Very clever.

src/language/sourceInstaller.ts
```diff
@@ -0,0 +1,127 @@
...
+import { promisify } from 'util';
+import { exec } from 'child_process';
+import * as fs from 'fs';
+import { StandaloneLanguageServerInstaller } from './standaloneLanguageServerInstaller';
+const execAsync = promisify(exec);
+const mkdirAsync = promisify(fs.mkdir);
+
+export class SourceInstaller {
+  public constructor(
+    private readonly oldSkoolInstaller: StandaloneLanguageServerInstaller
```
MikaelMayer, line 19: hum hum.... 😸 
```suggestion
    private readonly installerFromSource: StandaloneLanguageServerInstaller
```

src/language/standaloneLanguageServerInstaller.ts
```diff
@@ -0,0 +1,219 @@
...
+      const installed = await this.install();
+      if(!installed) {
+        throw new Error(Messages.Installation.Error);
+      }
+    }
+
+    return { command: dotnetExecutable, args: [ dllPath, ...args ] };
+  }
+
+  private async install(): Promise<boolean> {
```
MikaelMayer, line 55: How did you ensure this works the same way? Can you at least give it a try?

MikaelMayer (changes requested):
I can't wait to use the new version ! Let's agree on the next steps, implement it and get it merged.

package.json
```diff
@@ -115,11 +115,6 @@
       "type": "object",
       "title": "Dafny extension configuration",
       "properties": {
-        "dafny.languageServerRuntimePath": {
-          "type": "string",
-          "default": null,
-          "description": "Absolute or relative path to the DafnyLanguageServer.dll"
-        },
```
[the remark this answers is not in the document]
keyboardDrummer: You can use previous versions of Dafny by selecting them from the list. When using a custom Dafny, it has to be 3.10 or newer.

src/language/dafnyInstallation.ts
keyboardDrummer: Well.. except it doesn't work well since Z3 is not provided. Leaving this code unused for now.

src/language/standaloneLanguageServerInstaller.ts
keyboardDrummer: I will. It's in the testing list although still has a TODO.

package.json
```diff
@@ -184,12 +179,12 @@
...
           "default": [],
-          "description": "Optional array of strings to pass to the language server runtime."
+          "description": "Optional array of strings to pass to the language server."
         },
-        "dafny.compilerRuntimePath": {
+        "dafny.cliPath": {
           "type": "string",
           "default": null,
-          "description": "Absolute or relative path to the Dafny.dll"
+          "description": "Absolute or relative path to the Dafny CLI. When configured, overrides the selected Dafny version."
```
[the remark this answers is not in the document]
keyboardDrummer: I like your suggestion of having 'custom' in the dropdown. I'll do it here.

package.json
```diff
@@ -184,12 +179,12 @@
             "type": "string"
           },
           "default": [],
-          "description": "Optional array of strings to pass to the language server runtime."
+          "description": "Optional array of strings to pass to the language server."
         },
-        "dafny.compilerRuntimePath": {
+        "dafny.cliPath": {
```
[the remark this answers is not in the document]
keyboardDrummer: But now it's used both for the language server and the CLI. How would we merge the two previously configured settings into one?

package.json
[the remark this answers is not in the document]
MikaelMayer: Can you please point me to the code that will actually run the language server DLL with the language server options, if you select a previous version of Dafny? I must have missed it in the code review.

src/language/dafnyInstallation.ts
MikaelMayer: Oh no! Another great reason to want to depend on Z3 directly.

package.json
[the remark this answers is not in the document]
MikaelMayer: HMmm I guess it's even less important to keep the same name, now that users will have to select "custom" anyway in the dropdown list. I'm ok with the name change now.

src/language/customPathInstaller.ts
```diff
@@ -0,0 +1,152 @@
...
+      }
+      window.showInformationMessage(`Using $DAFNY_SERVER_OVERRIDE = ${cliPathOverride} for the CLI path`);
+    }
+    if(cliPathOverride) {
+      return cliPathOverride;
+    }
+
+    const version = getPreferredVersion();
+    if(version !== LanguageServerConstants.Custom) {
+      return '';
```
MikaelMayer, line 73: This method never returned an empty string before. Why is it doing that now?

src/language/customPathInstaller.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    if(server) {
+      newArgs.unshift('server');
+    }
+    if(configuredCliPath.endsWith('.dll')) {
+      const { path: dotnetExecutable } = await getDotnetExecutablePath();
+      return {
+        command: dotnetExecutable,
+        args: [ configuredCliPath ].concat(newArgs)
```
MikaelMayer, line 39: I thought that the newArgs for the DLL were different than the newArgs for the dafny server command. How come you can use them here?

src/language/dafnyInstallation.ts
```diff
@@ -1,163 +1,38 @@
...
-        cwd: executable.options?.cwd
-      });
-      return true;
-    } catch(e: unknown) {
-      window.showErrorMessage(`Could not start the Dafny CLI using ${JSON.stringify(executable)} because '${e}'. Please check if the installation is corrupt.`);
-      return false;
+    if(os.type() === 'Darwin' && os.arch() !== 'x64') {
+      // Need to build from source and move all files from Binary/ to the out/resource folder
+      const sourceInstaller = new FromSourceInstaller(this.context, this.statusOutput);
+      return await sourceInstaller.getExecutable(server, newArgs, oldArgs);
```
MikaelMayer, line 30: Very interesting refactoring, I like it.

MikaelMayer (commented):
A few more comments and questions

src/language/customPathInstaller.ts
keyboardDrummer: Both are using the Dafny server command, but one is using a standalone executable and the other a .dll that is run using dotnet

package.json
```diff
@@ -192,31 +192,28 @@
...
           },
           "default": [],
-          "description": "Optional array of strings to pass to the language server."
+          "description": "Optional array of strings to pass to the language server. Only works with Dafny versions above 3.10"
         },
         "dafny.cliPath": {
           "type": "string",
           "default": null,
-          "description": "Absolute or relative path to the Dafny CLI. When configured, overrides the selected Dafny version."
+          "description": "Path to the Dafny CLI. Only used when 'custom' is the chosen Dafny version. Dafny version must be at least 3.10"
```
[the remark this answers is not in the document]
keyboardDrummer: I left out the dot intentionally because it might be interpreted as part of the version number. What do you think? I can go either way.

package.json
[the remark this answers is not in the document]
keyboardDrummer: Here it uses `oldArgs`: https://github.com/dafny-lang/ide-vscode/pull/314/files#diff-a6fd99c63bc730191654b24b38cd9b5d7e3353548253e322a4271488ca29b7dfR55

src/language/customPathInstaller.ts
keyboardDrummer: For strings it's safest to use an empty string as the "not set" value, since UI components will return '' when users clear the field. It can also return empty string on line 79 where it returns the UI value.

package.json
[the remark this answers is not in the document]
MikaelMayer: I hear and share your concern. That's a parsing ambiguity. Let's solve it this way:
"The Dafny version must be 3.10 or greater."

package.json
```diff
@@ -192,7 +192,7 @@
             "type": "string"
           },
           "default": [],
-          "description": "Optional array of strings to pass to the language server. Only works with Dafny versions above 3.10"
+          "description": "Optional array of strings to pass to the language server. Only works with Dafny versions above 3.10. Available options can be seen using 'dafny server --help'"
```
MikaelMayer, line 195: That's a great idea. Because some users might not have Dafny in their code path, could you perhaps also provide a link to the document reference about these options?

src/language/githubReleaseInstaller.ts
```diff
@@ -73,7 +73,6 @@ export class GitHubReleaseInstaller {
     } catch(error: unknown) {
       this.writeStatus('Standalone language server installation failed:');
       this.writeStatus(`> ${error}`);
-      console.error('dafny installation failed', error);
```
MikaelMayer: I think this is a good summary of handling any error. Could you put it back please or justify here why you removed it?

src/test/suite/extension.test.ts
```diff
@@ -77,7 +90,7 @@ suite('Compiler invocation', () => {
     for(const checkpoint in checkpoints) {
       assert.strictEqual(true, (checkpoints as any)[checkpoint], checkpoint);
     }
-    assert.strictEqual('<compiler command prefix>"<dotnet executable path>" "<extension path>/<compiler runtime path>" "fileName.dfy" /out <arg1> arg2',
+    assert.strictEqual('cd <dotnet tool path>; <compiler command prefix>"<dotnet executable path>" build --output <arg1> arg2 "fileName.dfy"',
```
MikaelMayer, line 93: Ahah it's great to have tests to see what the tool does now ^^.

src/test/suite/extension.test.ts
```diff
@@ -90,7 +103,7 @@ suite('Dafny IDE Extension Installation', () => {
     assert.strictEqual(true, Messages.Dotnet.DownloadUri.includes('https'), 'Download URL should contain https');
   });
 
-  test('Installer checks', async () => {
+  test.only('Installer checks', async () => {
```
MikaelMayer, line 106: Did you forget this .only?

MikaelMayer (commented):
Requesting a few edits, next time I think we can merge !

package.json
keyboardDrummer: We don't have such a document reference yet. .NET as command specific documentation pages, like here: https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-build, but we don't have that.

src/language/githubReleaseInstaller.ts
keyboardDrummer: It ends up being printed while running the tests which is noisy. Is it important to print to the console given that it's also printed to the output channel?

package.json
MikaelMayer: Could you add the documentation of `dafny server --help` in the Dafny help, in another PR? You could use the same mechanism as "options.txt" to keep this help up to date. I think it would be really helpful, as we expect new users to not want to install Dafny manually to run this command, yet have access to the documentation of all the options they can now configure on VSCode thanks to your PRs.
Here is a good place to add this documentation. Perhaps it needs some refactoring.
https://dafny.org/latest/DafnyRef/DafnyRef#sec-dafny-language-server-vscode

src/language/githubReleaseInstaller.ts
MikaelMayer: I hadn't realized it was console.error(). I guess it's fine then since it's not displayed to the user.

MikaelMayer (approved):
Looking forward to use it! Thanks for putting it up.

package.json
keyboardDrummer: @davidcok is this something you could incorporate in https://github.com/dafny-lang/dafny/pull/3220 ?

PR #316, opened 2022-12-15
Always attempt to find a binary distribution

atomb:
This updates the installation process to always try to download a binary distribution, using the current OS _and_ architecture, and only build from source if the download fails. This should allow us to support other architectures seamlessly in the future (and support the case where, for some reason, we fail to upload a binary for a certain version/OS/arch combination).

### DIFF
--- a/src/language/dafnyInstallation.ts
+++ b/src/language/dafnyInstallation.ts
@@ -265,7 +265,7 @@ export class DafnyInstaller {
       console.error('dafny download failed, trying to build from source', error);
       try {
         // Need to build from source and move all files from Binary/ to the out/resource folder
-        this.writeStatus(`Failed to download binary distribution of Dafny. Installing from source instead.`);
+        this.writeStatus('Failed to download binary distribution of Dafny. Installing from source instead.');
         return await this.installFromSource();
       } catch(error: unknown) {
         this.writeStatus('Dafny build failed:');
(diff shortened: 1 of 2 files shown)


src/language/dafnyInstallation.ts
```diff
@@ -254,22 +256,25 @@ export class DafnyInstaller {
...
-        this.writeStatus(`Found a non-supported architecture OSX:${os.arch()}. Going to install from source.`);
+        this.writeStatus('Failed to download binary distribution of Dafny. Installing from source instead.');
         return await this.installFromSource();
-      } else {
-        const archive = await this.downloadArchive(await getDafnyDownloadAddress(this.context), 'Dafny');
-        await this.extractArchive(archive, 'Dafny');
-        await workspace.fs.delete(archive, { useTrash: false });
-        this.writeStatus('Dafny installation completed');
-        return true;
+      } catch(error: unknown) {
```
MikaelMayer, line 272: The problem is that the install from source is only available for mac os users, because it uses the command `await this.execLog('brew install dotnet-sdk');`
Could you either ensure that non-mac users can't install from source, or add OS detection and install dotnet another way if necessary?

atomb: Oh, good point. I tested this, but only on macOS. :) I'll add a further check.

keyboardDrummer (approved):
Amazing changes, thanks!

PR #323, opened 2022-12-21
Update package.json

MikaelMayer:
This PR provides a better hint so that users checking out an older version of Dafny still know that they can use the path to the language server.

Is this correct @keyboardDrummer ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -192,12 +192,12 @@
             "type": "string"
           },
           "default": [],
-          "description": "Optional array of strings to pass to the language server. Only works with Dafny versions above 3.10. Available options can be seen using 'dafny server --help'"
+          "description": "Optional array of strings to pass to the language server. Only works if the Dafny version is 3.10 or greater. Available options can be seen using 'dafny server --help'"
         },
         "dafny.cliPath": {
           "type": "string",
           "default": null,
-          "description": "Path to the Dafny CLI. Only used when 'custom' is the chosen Dafny version. The Dafny version must be 3.10 or greater."
+          "description": "Path to Dafny.exe, Dafny.dll or DafnyLanguageServer.dll. Only used when 'custom' is the chosen Dafny version."
         },
         "dafny.compilerOutputDir": {
           "type": "string",
--- a/package.json
+++ b/package.json
@@ -197,7 +197,7 @@
         "dafny.cliPath": {
           "type": "string",
           "default": null,
-          "description": "Path to Dafny.exe or Dafny.dll. Only used when 'custom' is the chosen Dafny version."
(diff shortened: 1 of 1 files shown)


keyboardDrummer (approved):
~Good change, thanks.~

package.json
```diff
@@ -192,12 +192,12 @@
...
           },
           "default": [],
-          "description": "Optional array of strings to pass to the language server. Only works with Dafny versions above 3.10. Available options can be seen using 'dafny server --help'"
+          "description": "Optional array of strings to pass to the language server. Only works if the Dafny version is 3.10 or greater. Available options can be seen using 'dafny server --help'"
         },
         "dafny.cliPath": {
           "type": "string",
           "default": null,
-          "description": "Path to the Dafny CLI. Only used when 'custom' is the chosen Dafny version. The Dafny version must be 3.10 or greater."
+          "description": "Path to Dafny.exe, Dafny.dll or DafnyLanguageServer.dll. Only used when 'custom' is the chosen Dafny version."
```
keyboardDrummer, line 200: I don't think it works with `DafnyLanguageServer.dll` or versions older than 3.10, since for custom it always uses the new type of CLI arguments. It might sort of work but without using all the preferences.

MikaelMayer: Could you please at least test it and ensure it works? Without being able to use DafnyLanguageServer.dll it would be difficult, for example, to bissect the language server to detect a regression issue. There might be workarounds by using nightly releases but I wouldn't rely on them as they might be overriden.

keyboardDrummer: I tested it and it works but indeed only with the default options, so the preferences are ignored.

package.json
```diff
@@ -192,12 +192,12 @@
...
           },
           "default": [],
-          "description": "Optional array of strings to pass to the language server. Only works with Dafny versions above 3.10. Available options can be seen using 'dafny server --help'"
+          "description": "Optional array of strings to pass to the language server. Only works if the Dafny version is 3.10 or greater. Available options can be seen using 'dafny server --help'"
         },
         "dafny.cliPath": {
           "type": "string",
           "default": null,
-          "description": "Path to the Dafny CLI. Only used when 'custom' is the chosen Dafny version. The Dafny version must be 3.10 or greater."
+          "description": "Path to Dafny.exe or Dafny.dll. Only used when 'custom' is the chosen Dafny version."
```
keyboardDrummer, line 200: `Only works if the Dafny version is 3.10 or greater.` should be in there

keyboardDrummer, line 200: It doesn't need to be `Dafny.exe` or `Dafny.dll`, can also be a `dafny` UNIX executable file.

keyboardDrummer (approved):

PR #332, opened 2022-12-27
Introduce a temporary fix for verification cores set to 0

keyboardDrummer:
Introduce a temporary fix so setting verification cores to 0 (the default) doesn't stop verification


MikaelMayer (approved):
I'll approve it because it's urgent but I would like to have a follow-up PR with a test to ensure that, if the number of cores is not specified, that the command line used to run DafnyLanguageServer is not set to zero cores.

keyboardDrummer:
> I'll approve it because it's urgent but I would like to have a follow-up PR with a test to ensure that, if the number of cores is not specified, that the command line used to run DafnyLanguageServer is not set to zero cores.

Note that we're also updating the language server to allow 0 cores: https://github.com/dafny-lang/dafny/pull/3276

PR #334, opened 2022-12-27
Reinstated the copy of the executable

MikaelMayer:
With 3.0.1, the Dlls and executables were no longer copied from a custom installation to the new one. This prevents Dafny developers to compile and then refresh VSCode as it was enabled recently.

This PR fixes that.

### DIFF
--- a/src/language/customPathInstaller.ts
+++ b/src/language/customPathInstaller.ts
@@ -56,43 +56,42 @@ export class CustomPathInstaller {
 
   private async getCliPathUncached(context: ExtensionContext): Promise<string> {
     let cliPathOverride = process.env['DAFNY_SERVER_OVERRIDE'] ?? '';
-    if(cliPathOverride) {
+    const version = getPreferredVersion();
+    let cliPath = Configuration.get<string | null>(ConfigurationConstants.LanguageServer.CliPath) ?? '';
+    if(cliPath && !path.isAbsolute(cliPath)) {
+      cliPath = path.join(context.extensionPath, cliPath);
+    }
+    if(cliPathOverride || cliPath && version === LanguageServerConstants.Custom) {
+      const originalPath = cliPathOverride || cliPath;
+      let acopyofDafny = originalPath;
       try {
-        cliPathOverride = await this.copyNecessaryDllsToTempFolder(cliPathOverride);
+        cliPathOverride = await this.copyNecessaryDllsToTempFolder(originalPath);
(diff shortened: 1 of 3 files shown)


src/language/customPathInstaller.ts
```diff
@@ -56,43 +56,42 @@ export class CustomPathInstaller {
 
   private async getCliPathUncached(context: ExtensionContext): Promise<string> {
     let cliPathOverride = process.env['DAFNY_SERVER_OVERRIDE'] ?? '';
-    if(cliPathOverride) {
+    const version = getPreferredVersion();
+    let cliPath = Configuration.get<string | null>(ConfigurationConstants.LanguageServer.CliPath) ?? '';
```
keyboardDrummer, line 60: Why would we also let the configured custom language server cause a dll copy, instead of only when using the Dafny language developer option `DAFNY_SERVER_OVERRIDE` ? I thought the dll copying was for Dafny language developers.

MikaelMayer: I am a Dafny language developer and I am not using DAFNy_SERVER_OVERRIDE because I'm never launching vs code from the command line. In my case, I always start VSCode from scratch, open the extension repository, and then launch a second version of VSCode from there to test Dafny. I don't know how to set up this environment variable - and I probably don't want to, because I regularly switch between Dafny versions when encountering a bug (and I simply hit CTRL+R to reload the second window).
Moreover, the behavior previously was to copy all the dlls and executables both in the case of DAFNY_SERVER_OVERRIDE and in the case of a custom installation. Why did you change that in the first place?

Now to the point: Custom installations are often recompiled, and reading the dlls and executables directly prevents the compilation to go through - meaning one has to close VSCode, compile and relaunch vscode.
Not copying the dlls is surely faster, but when we develop, it's a performance bottleneck - @robin-aws  had the same one and can testify.

keyboardDrummer: > Why did you change that in the first place?

I assumed anyone compiling Dafny from source would use `DAFNY_SERVER_OVERRIDE`. Sorry for that.

> I am a Dafny language developer and I am not using DAFNy_SERVER_OVERRIDE because I'm never launching vs code from the command line. In my case, I always start VSCode from scratch, open the extension repository, and then launch a second version of VSCode from there to test Dafny. I don't know how to set up this environment variable - and I probably don't want to, because I regularly switch between Dafny versions when encountering a bug (and I simply hit CTRL+R to reload the second window).

Makes sense.

keyboardDrummer (approved):

PR #336, opened 2022-12-28
Fix non-integer number of cores.

mzhu65536-um:
DafnyLanguageServer crashes because the way the number of cores is computed could result in a float number where the server option is expecting an `uint`.

> Cannot parse argument '8.5' for option '--cores' as expected type 'System.Int32'.


MikaelMayer (approved):
Wow sounds like an important fix.

PR #347, opened 2023-01-24
Prepare migration for Dafny 4

MikaelMayer:
This PR adds a feedback mechanism is users using "Latest Stable" or "Latest Nightly" one day encounter a major version bump.
If so, it displays a dialog box like this:

image

### DIFF
--- a/src/language/dafnyInstallation.ts
+++ b/src/language/dafnyInstallation.ts
@@ -1,5 +1,5 @@
 /* eslint-disable max-depth */
-import { ExtensionContext, OutputChannel } from 'vscode';
+import { ExtensionContext, OutputChannel, env, Uri, window } from 'vscode';
 import { ConfigurationConstants } from '../constants';
 import Configuration from '../configuration';
 import { Executable } from 'vscode-languageclient/node';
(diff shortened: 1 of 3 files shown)


src/language/githubReleaseInstaller.ts
```diff
@@ -121,32 +121,88 @@ export class GitHubReleaseInstaller {
     return this.getConfiguredTagAndVersionCache;
   }
 
+  public static readonly CurrentVersionTag = 'current-version';
+
+  // If there is a major version bump and the version is auto-selected
+  // warn the user before hand
+  private static async dafny4upgradeCheck(
```
keyboardDrummer, line 128: What's the code duplication for?

MikaelMayer: I wanted to check with you, it looks like this one is the only active place, I thought you had activated the dotnet tool installer, but apparently the class is not used anywhere. Is that intended? I plan to either remove one or reduce duplication.

src/language/dafnyToolInstaller.ts
```diff
@@ -59,12 +59,51 @@ class DafnyToolInstaller {
...
+          `Dafny ${newToolVersion} is out! You are using the ${versionDescription.toLowerCase()}, `
+          + 'but because this is a major change, we want you to confirm the switch.',
+          seeMigrationInstructions,
+          accept,
+          decline
+        );
+        if(answer === undefined || answer === decline) {
+          newToolVersion = (compareToolVersion !== undefined ? compareToolVersion : newToolVersion) as string;
+        } else if(answer === seeMigrationInstructions) {
+          env.openExternal(Uri.parse('https://dafny.org/TODO'));
```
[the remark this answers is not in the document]
MikaelMayer: Ideally, we want to have a link before. @davidcok , would you have such a link handy, at least temporary?

src/language/githubReleaseInstaller.ts
keyboardDrummer: The tool installer isn't active because it doesn't include Z3 and the tool doesn't know how to download Z3.

MikaelMayer: Let me just refactor it for now so that it's not duplicated, but we'll keep it there just in case.

keyboardDrummer (approved):

src/language/dafnyInstallation.ts
```diff
@@ -18,6 +18,47 @@ export class DafnyInstaller {
...
+          `Dafny ${newToolVersion} is out! You are using the ${versionDescription.toLowerCase()}, `
+          + 'but because this is a major change, we want you to confirm the switch.',
+          seeMigrationInstructions,
+          accept,
+          decline
+        );
+        if(answer === undefined || answer === decline) {
+          newToolVersion = (compareToolVersion !== undefined ? compareToolVersion : newToolVersion) as string;
+        } else if(answer === seeMigrationInstructions) {
+          env.openExternal(Uri.parse('https://dafny.org/TODO'));
```
keyboardDrummer, line 52: Why not refer to the release notes?

keyboardDrummer (changes requested):
The `'https://dafny.org/TODO'` is still in there so I can't approve. Do we even need to provide a migration instructions link?

src/language/dafnyInstallation.ts
MikaelMayer: Our release notes won't be detailed enough for that purpose. We need a better migration document. Perhaps @davidcok  knows where would be a good place to start such a document?

MikaelMayer:
> The `'https://dafny.org/TODO'` is still in there so I can't approve. Do we even need to provide a migration instructions link?

Yes we need to provide a migration instructions link. Consider a usual Dafny user who hasn't heard about all the things that will be different in Dafny 4. Some error messages will just make no sense to them "Hey this is a ghost function, why are you complaining Dafny? Uh oh, Dafny is broken with the last version, let me cut an issue.".

keyboardDrummer (approved):

PR #348, opened 2023-01-30
Add verification integration test

keyboardDrummer:
Add integration test that checks that verification through a solver works


MikaelMayer (approved):
That's a great start to ensure VSCode does not break on an update ! Thanks for investigating how to do it and actually do it.

PR #349, opened 2023-02-01
Release v3.0.4 (updated Dafny to 3.11.0)

keyboardDrummer:
- Update version to 3.0.4
- Use Dafny 3.11
- Remove the major version upgrade checks when 'latest nightly' is used, since they would always trigger because the 'version' for nightly starts with `n` and that's never equal to the latest major version number.
- Remove `.only` from test, so the other tests run again

### DIFF
--- a/src/test/verification.test.ts
+++ b/src/test/verification.test.ts
@@ -1,70 +1,8 @@
 import * as assert from 'assert';
 import * as vscode from 'vscode';
-import { ViewColumn } from 'vscode';
 
 suite('Verification', () => {
-  test('Program with errors has diagnostics', async () => {
-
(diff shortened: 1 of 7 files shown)


stefan-aws (approved):
Looks good

src/test/verification.test.ts
```diff
@@ -2,8 +2,7 @@ import * as assert from 'assert';
 import * as vscode from 'vscode';
 
 suite('Verification', () => {
-  test.only('Program with errors has diagnostics', async () => {
-
+  test('Program with errors has diagnostics', async () => {
```
robin-aws, line 5: Oof, good catch. Perhaps for later (not blocking this PR) we should add a CI check to prevent this from getting checked in.

robin-aws (approved):

PR #356, opened 2023-02-21
fix: check commit msg's first line for lastPreparedTag

alex-chew:
This avoids the edge case where a commit message has a semver-style version number in the message body (as is often the case in dependabot commits), which results in the publish process erroneously exiting early.

### DIFF
--- a/publish_process.js
+++ b/publish_process.js
@@ -171,7 +171,7 @@ async function nextVersion(currentVersion) {
 }
 
 async function getLastPreparedTag() {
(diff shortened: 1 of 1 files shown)


davidcok (approved):
Looks OK as long as execAsync takes a pipe of shell commands

alex-chew:
Good call, and indeed it does as per the documentation's examples: https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback

PR #361, opened 2023-02-28
Update the names of release packages

atomb:
For 3.13 and later, the names of the binary packages we produce has changed.

### DIFF
--- a/src/language/githubReleaseInstaller.ts
+++ b/src/language/githubReleaseInstaller.ts
@@ -15,18 +15,29 @@ import { DafnyInstaller, getPreferredVersion } from './dafnyInstallation';
 import { configuredVersionToNumeric } from '../ui/dafnyIntegration';
 const ArchiveFileName = 'dafny.zip';
 
-function getDafnyPlatformSuffix(): string {
+function getDafnyPlatformSuffix(version: string): string {
+  const post312 = version.includes('nightly') || configuredVersionToNumeric(version) >= configuredVersionToNumeric('3.13');
   switch(os.type()) {
   case 'Windows_NT':
-    return 'win';
+    if(post312) {
+      return 'windows-2019';
+    } else {
+      return 'win';
+    }
   case 'Darwin':
-    if(os.arch() === 'arm64') {
+    if(post312) {
+      return 'macos-11';
+    } else if(os.arch() === 'arm64') {
       return 'osx-11.0';
     } else {
       return 'osx-10.14.2';
     }
   default:
-    return 'ubuntu-16.04';
+    if(post312) {
+      return 'ubuntu-20.04';
+    } else {
+      return 'ubuntu-16.04';
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/language/githubReleaseInstaller.ts
```diff
@@ -15,18 +15,29 @@ import { DafnyInstaller, getPreferredVersion } from './dafnyInstallation';
 import { configuredVersionToNumeric } from '../ui/dafnyIntegration';
 const ArchiveFileName = 'dafny.zip';
 
-function getDafnyPlatformSuffix(): string {
+function getDafnyPlatformSuffix(version: string): string {
+  const post312 = version.includes('nightly') || configuredVersionToNumeric(version) >= configuredVersionToNumeric('3.13');
   switch(os.type()) {
   case 'Windows_NT':
-    return 'win';
+    if(post312) {
```
robin-aws, line 22: This seems like it would be cleaner if the `case` was inside a single `if` instead, since all three cases are different post 3.12

src/language/githubReleaseInstaller.ts
```diff
@@ -15,18 +15,29 @@ import { DafnyInstaller, getPreferredVersion } from './dafnyInstallation';
 import { configuredVersionToNumeric } from '../ui/dafnyIntegration';
 const ArchiveFileName = 'dafny.zip';
 
-function getDafnyPlatformSuffix(): string {
+function getDafnyPlatformSuffix(version: string): string {
+  const post312 = version.includes('nightly') || configuredVersionToNumeric(version) >= configuredVersionToNumeric('3.13');
```
robin-aws, line 19: The 'nightly' part is slightly wrong, since in the context of this method it could be an older nightly build. If that's not actually a concern because of what this function is called on, we should add a comment.

robin-aws (changes requested):
Is there any way to test this ahead of time?

src/language/githubReleaseInstaller.ts
atomb: Yeah, perhaps so.

src/language/githubReleaseInstaller.ts
atomb: Since it always fetches the most recent nightly, I don't think it can fail. I'll add a comment about that.

atomb:
> Is there any way to test this ahead of time?

I've tested locally that it correctly downloads the latest nightly build. I can't test it with the 3.13 or 4.0 releases yet, of course. :)

src/language/githubReleaseInstaller.ts
```diff
@@ -15,18 +15,33 @@ import { DafnyInstaller, getPreferredVersion } from './dafnyInstallation';
...
+  if(post312) {
+    switch(os.type()) {
+    case 'Windows_NT':
+      return 'windows-2019';
+    case 'Darwin':
+      return 'macos-11';
+    default:
+      return 'ubuntu-20.04';
+    }
```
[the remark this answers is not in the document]
atomb: Yeah. We could decide to leave out the versions (in a later change, so we'd have three cases). They're often useful to judge compatibility when downloading manually, but not otherwise necessary.

PR #363, opened 2023-03-02
Update README.md

jtristan:


MikaelMayer (approved):
We really should fix the script publish_process so that we don't need to do this in the future

PR #369, opened 2023-03-06
Fix undesired red bracket

### LINKED ISSUE #368: Undesired red highlights
Since a recent upgrade of VSCode (not the extension), some right brackets turn red.

image

This is undesirable

MikaelMayer:
Fixes #368


robin-aws (commented):
It looks like that entry has been there for a long time, what changed?

MikaelMayer:
> It looks like that entry has been there for a long time, what changed?

I have to trace for VSCode's history, but apparently someone thought it would be clever to highlight in red unmatched brackets without providing a way to prevent this.
Note: previously, in the expression `i < x == j > y`, if you had placed the cursor over the first `<`, it would have highlighted the second `>`, which obviously has never been correct but it did not bother any of us ^^

robin-aws (approved):

PR #377, opened 2023-03-20
Support for highlighting the keyword "opaque"

MikaelMayer:


src/test/highlight/regression-highlight.dfy
```diff
@@ -1,3 +1,7 @@
+// Support for opaque keyword
+opaque function Test(): int { 1 }
+//^^^^ should be hihglighted as a keyword
```
davidcok, line 3:
```suggestion
//^^^^ should be highlighted as a keyword
```

syntaxes/Dafny.tmLanguage
```diff
@@ -26,7 +26,7 @@
 				</dict>
 			</dict>
 			<key>end</key>
-			<string>:=|:\||;|\)|\}|(?=(?:class|codatatype|const|constructor|datatype|function|greatest|iterator|least|lemma|method|module|newtype|predicate|trait|type|var)\b)</string>
+			<string>:=|:\||;|\)|\}|(?=(?:class|codatatype|const|constructor|datatype|function|greatest|iterator|least|lemma|method|module|newtype|opaque|predicate|trait|type|var)\b)</string>
```
davidcok, line 29: If this is intended to be a full list of keywords, there are some missing. abstract, static, ghost are in the same category (modifiers) as opaque

syntaxes/Dafny.tmLanguage
```diff
@@ -142,7 +142,7 @@
 				</dict>
 			</dict>
 			<key>end</key>
-			<string>(?=&lt;|\s*|,|\)|\}|(?:requires|ensures|modifies|reads|class|codatatype|const|constructor|datatype|function|greatest|iterator|least|lemma|method|module|newtype|predicate|trait|type|var)\b))</string>
+			<string>(?=&lt;|\s*|,|\)|\}|(?:requires|ensures|modifies|reads|class|codatatype|const|constructor|datatype|function|greatest|iterator|least|lemma|method|module|newtype|opaque|predicate|trait|type|var)\b))</string>
```
davidcok, line 145: ditto

syntaxes/Dafny.tmLanguage
```diff
@@ -448,7 +448,7 @@
 			<array>
 				<dict>
 					<key>match</key>
-					<string>\b(class|trait|function|include|ghost|nameonly|var|const|method|constructor|abstract|module|import|export|lemma|as|is|opened|static|twostate|refines|witness|extends|returns|break|then|else|if|label|return|while|for|to|downto|print|new|in|fresh|allocated|match|case|assert|by|assume|expect|reveal|modify|predicate|least|greatest|forall|exists|false|true|null|old|unchanged|calc|iterator|yields|yield)\b</string>
+					<string>\b(class|trait|function|include|ghost|nameonly|var|const|method|constructor|abstract|module|import|export|lemma|as|is|opened|static|twostate|refines|witness|extends|returns|break|then|else|if|label|return|while|for|to|downto|print|new|in|fresh|allocated|match|case|assert|by|assume|expect|reveal|modify|predicate|least|greatest|forall|exists|false|true|null|old|unchanged|calc|iterator|opaque|yields|yield)\b</string>
```
davidcok, line 451: If these were in alphabetical order, it would help to see what is missing. For example, I see 'break' but not 'continue'

davidcok (changes requested):

syntaxes/Dafny.tmLanguage
MikaelMayer:
```suggestion
			<string>:=|:\||;|\)|\}|(?=(?:class|codatatype|const|constructor|datatype|function|greatest|iterator|least|lemma|method|module|newtype|opaque|predicate|trait|type|var|abstract|static|ghost)\b)</string>
```

syntaxes/Dafny.tmLanguage
MikaelMayer:
```suggestion
			<string>(?=&lt;|\s*|,|\)|\}|(?:requires|ensures|modifies|reads|class|codatatype|const|constructor|datatype|function|greatest|iterator|least|lemma|method|module|newtype|opaque|predicate|trait|type|var|abstract|static|ghost)\b))</string>
```

syntaxes/Dafny.tmLanguage
MikaelMayer:
```suggestion
					<string>\b(abstract|allocated|as|assert|assume|break|by|calc|case|class|const|constructor|continue|downto|else|exists|expect|export|extends|false|for|forall|fresh|function|ghost|greatest|if|import|in|include|is|iterator|label|least|lemma|match|method|modify|module|nameonly|new|null|old|opaque|opened|predicate|print|refines|return|returns|reveal|static|then|to|trait|true|twostate|unchanged|var|while|witness|yield|yields)\b</string>
```

syntaxes/Dafny.tmLanguage
MikaelMayer: Here as well, it's only a list of keywords to stop the parsing of variable: type

MikaelMayer:
I tested again on Highlights.dfy and found one issue, I had accidentally deleted the "var" keyword to recognize that a variable declaration stopped

alex-chew (dismissed):

davidcok (approved):

PR #392, opened 2023-05-19
Support new IDE states, parsing and preparing verification

keyboardDrummer:
### Testing

- New statuses are shown when using Dafny master
- Statuses still work as they did when using Dafny 4.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -287,6 +287,8 @@
       }
     ]
   },
+  "activationEvents": [
+  ],
   "scripts": {
     "vscode:prepublish": "npm run package",
     "compile": "webpack",
--- a/package.json
+++ b/package.json
@@ -287,9 +287,6 @@
       }
     ]
   },
-  "activationEvents": [
-    "onLanguage:dafny"
-  ],
   "scripts": {
     "vscode:prepublish": "npm run package",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -287,9 +287,6 @@
       }
     ]
   },
-  "activationEvents": [
```
keyboardDrummer, line 290: This is inferred from the language being added in the package.json

src/ui/compilationStatusView.ts
```diff
@@ -36,6 +40,7 @@ function toStatusMessage(status: CompilationStatus, message?: string | null): st
       ? `${Messages.CompilationStatus.VerificationFailedOld} ${message}`
       : `${Messages.CompilationStatus.VerificationFailedOld}`;
   }
+  return status.toString();
```
MikaelMayer, line 43: This is for future forward compatibility, rivht?

MikaelMayer (approved):
Thanks for the follow up !

src/ui/compilationStatusView.ts
keyboardDrummer: I don't think it's needed for the current changes, but I thought it might be good defensive programming

Issue #378, opened 2023-03-31
Change the order of offered Dafny versions


MikaelMayer:
Currently Dafny offers versions like:

```
latest                  default
4.0.0
3.13.1
3.12.0
3.11.0
3.10.0
3.9.1
3.9.0
3.8.1
3.8.0
3.7.3
3.6.0
3.5.0
3.4.2
3.3.0
3.2.0
custom
latest-nightly
```

There are several problems with this order:
1. Everything below 3.8.1 is invisible until we scroll
2. It's unclear what `latest` means. Is it `4.0.0` or is it a more recent version?
3. Nobody sees `latest-nightly` when they find this field, it's quite unexpected that it's at the end of the list, one would expect only to see. We did that because we wanted everyone to use the latest stable release, not the latest nightly by default. However, what the default is is correclty highlighted by VSCode already (see 'default')
4. "custom" is between two versions of Dafny.
5. There are too many Dafny versions.

Here what I suggest:
* Use  local (uses the Cli Path option)` instead of `custom` and put it above everything instead of at the end to indicate that it's probably going to be the most up to date if selected, and what it actually requires so that it's not ambiguous. 'custom' would indicate that we would be able to specify the version, which we cannot.
* Put `latest-nightly` above `4.0.0` to indicate it's even more recent
* Replace `latest-nightly` with the more verbose `always latest nightly release`
* Replace `latest` by `always latest release` with between parentheses `currently <the latest Dafny version>`.
* Remove all versions and replace them by `custom release (uses the Fixed Custom Version option)`
* Add an option `Custom Version` dropdown that contains all the released versions of Dafny.

Thus the list would look like the much simpler, it's possible to use the probable chronological order of versions, etc:

Dafny **Version**
The preferred Dafny version to use (requires restart)
```
local (uses the Cli Path option)
always latest nightly release
always latest release (currently 4.0.0)                  default
custom release (uses the Fixed Custom Release Version option)
```

Dafny **Fixed Custom Release Version**
A fixed Dafny release version to use (requires restart and Dafny Version to be set to `custom`)
```
4.0.0
3.13.1
3.12.0
3.11.0
3.10.0
3.9.1
3.9.0
3.8.1
3.8.0
3.7.3
3.6.0
3.5.0
3.4.2
3.3.0
3.2.0
```

PR #419, opened 2023-08-14
Clarify enum option in Dafny version picker

keyboardDrummer:


MikaelMayer (changes requested):
It's one step closer to #378, thanks for that.
Please make sure that you also update constants.ts

keyboardDrummer:
> It's one step closer to #378, thanks for that. Please make sure that you also update constants.ts

Yikes, thanks. I might have tested with a language server override, so the server worked regardless of this bug.

PR #420, opened 2023-08-14
Feat: Added the test command

### LINKED ISSUE #417: "Dafny > test"
I would like to just right-click and run Dafny > test, but this option is not present.
Currently, I select "build with custom arguments" and I replace everything with "test" or "test --no-verify".
it would be good to have the two options.
Test without verification is useful to check up if a property holds before proving it.

MikaelMayer:
Fixes #417
This PR adds the "Test" command in the contextual Dafny menu.

A bit of context. It's not possible to put several Main methods in included files, but it's definitely possible to put tests in each file as needed. Currently, to run tests, one typically right-clicks "Dafny > build with custom arguments", and then replace everything with "test". There is not even a shortcut for that, although the "test" command is fully supported in Dafny. Moreover, I am adding the "--no-verify" flag by default both on tests and run because 1) it's in the context of a development environment 2) The minimum required for `dafny run` and `dafnyt test` is that the program resolves and the compiler can compile it. 3) Verification of the entire file before testing is at best redundant because it was already verified in the IDE, at worst it forces the user to add a bunch of "assume false" in their code just because they haven't prove some assertions they conjectured.

I think it's fair to say that the usual workflow of testing is complementary to verification, meaning it makes it possible to experiment on specifications before proving them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,11 @@
           "group": "7_Dafny@7"
         },
         {
+          "when": "editorTextFocus && editorLangId == dafny",
+          "command": "dafny.test",
+          "group": "7_Dafny@8"
+        },
+        {
           "when": "editorTextFocus && editorLangId == dafny && dafny.counterexampleMenu.CanShowCounterExample",
           "command": "dafny.showCounterexample",
           "group": "7_Dafny@1"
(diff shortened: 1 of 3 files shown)


keyboardDrummer:
Commented on the issue: https://github.com/dafny-lang/ide-vscode/issues/417#issuecomment-1677667115

jtristan (approved):

PR #423, opened 2023-08-23
Client side gutter icons

keyboardDrummer:
### Changes
- Compute gutter icons on the client and configure the server so it no longer sends gutter icons to the client.

### Testing
- Manual testing looked good, but @MikaelMayer I don't know the icons well enough to know if it behaved the same. I think there might be differences in behavior when there is a resolution error, although the 'new' behavior seems fine to me.
- TODO I will add unit tests for the method `computeNewGutterIcons`, to check that given the particular symbolStatus, diagnostics and documentSymbol information, the right gutter icons are computed.
- I will not add an integration test that runs an actual language server and checks that the icons per line evolve according to expectations. Such a test would be difficult to make stable because it depends on having a running language server.

### DIFF
--- a/src/language/dafnyLanguageClient.ts
+++ b/src/language/dafnyLanguageClient.ts
@@ -34,7 +34,7 @@ function getLanguageServerLaunchArgsNew(): string[] {
     getVerifierCachingPolicy(),
     `--cores:${cores}`,
     `--notify-ghostness:${Configuration.get<string>(ConfigurationConstants.LanguageServer.MarkGhostStatements)}`,
-    `--notify-line-verification-status:${Configuration.get<string>(ConfigurationConstants.LanguageServer.DisplayGutterStatus)}`,
+    '--notify-line-verification-status:false',
     ...getDafnyPluginsArgument(),
     ...launchArgs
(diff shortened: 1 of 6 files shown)


MikaelMayer:
I just tried this PR with the latest Dafny master. I _typed_:

```dafny
method Test() {

}
```
and got the following result
image
Reloading VSCode on that same file displays nothing in gutter icons.
Even writing "assert true;" inside did not change anything.

The language server was started with these options.

```
Language server: {"command":"...\\AppData\\Local\\Temp\\vscode-dafny-dlls-W1YBH8\\Dafny.exe","args":["server","--verify-on:Change","--verification-time-limit:30","--cache-verification=0","--cores:4","--notify-ghostness:true","--notify-line-verification-status:false"]}
Dafny is ready
```

keyboardDrummer:
I'm in a branch that includes my unmerged PRs to Dafny server. Will let you know when those are merged.

Given the code you showed, when opening the file I get:

image

When pressing verification it turns into:
image


---
Different contents:

With manual verification, on opening file:
image

After clicking verify:
image

image

PR #436, opened 2023-09-20
Fix the removal of verification tests

keyboardDrummer:
### Changes
- Fix comparison so items in the the UI for verification tests are removed when they're out of date.

### Testing
- Manually tested that verification tests are now kept up to date


robin-aws (approved):
Nice, I noticed this recently myself

PR #475, opened 2024-05-08
Dafny project file improvements

keyboardDrummer:
### Changes
- No longer let the syntax definitions for .dfy files also be applied to dfyconfig.toml files
- Activate extension when `.toml` file is opened, such as a `dfyconfig.toml`
- Let the language client also operate on files ending in `dfyconfig.toml`

### Testing
- Manually tested that only opening a `dfyconfig.toml` will already trigger the extension and provide diagnostics.
- Manually tested that making changing in a `dfyconfig.toml` file causes diagnostics to be updated
- Manually tested that syntax highlighting for `dfyconfig.toml` files works correctly.


package.json
```diff
@@ -11,6 +11,9 @@
   "bugs": {
     "url": "https://github.com/dafny-lang/ide-vscode/issues"
   },
+  "activationEvents": [
+    "onLanguage:toml"
```
MikaelMayer, line 15: Isn't that a bit too broad? toml is not Dafny specific. This means that the Dafny extension will start even if a toml file is run that isn't a dfyconfig. Could you please create an empty .toml file and verify that it won't crash the language server somehow?

keyboardDrummer: Activating too early can not lead to bugs, but it hurts VSCode performance since the extension will trigger startup code even when you're just working with "*.toml" files that are unrelated to Dafny.

I would have preferred to set the extension to activate only when the path pattern "**/*dfyconfig.toml" is opened, but VSCode does not support that.

If I leave out this activation event, then if you open a `dfyconfig.toml` without opening any .dfy file, the extension does not activate. This is only a slight inconvenience, but starting the extension even when you're not doing anything with Dafny is also only slightly inconvenient.

MikaelMayer: Ok, I can live with that for now.

MikaelMayer (approved):

PR #483, opened 2024-07-09
Export ExtensionRuntime (or LanguageClient) for other extensions

BurstingF:
This change would export the `ExtensionRuntime` object that is created such that other extensions can access it.
This would, for example, allow other extensions to depend on this extension and use the `LanguageClient` it creates to send requests without the need to start their own `LanguageClient`.

With the current changes other extensions could access the `LanguageClient` like this:
```typescript
const client: LanguageClient = vscode.extensions.getExtension('dafny-lang.ide-vscode')?.exports.client;
client.sendRequest(...)
```

Note: Perhaps if exporting `ExtensionRuntime` is a bit too broad, only the `LanguageClient` could be exported instead.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -14,14 +14,15 @@ import * as PromiseAny from 'promise.any';
 const DafnyVersionTimeoutMs = 5_000;
 let extensionRuntime: ExtensionRuntime | undefined;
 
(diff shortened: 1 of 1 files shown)


keyboardDrummer (approved):

keyboardDrummer:
Seems good, thanks

PR #499, opened 2024-09-18
Update node used in CI

keyboardDrummer:


fabiomadge (approved):

.github/workflows/ci.yaml
```diff
@@ -110,9 +110,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: "Prepare: Use Node.js ${{ matrix.node-version }}"
```
fabiomadge, line 112: This doesn't seem to be necessarily true.

Issue #515, opened 2025-01-02
starting Dafny VSCode IDE on a MAC needs to be online or the verification can't start


kjx:
Failing code
------------
N/A

Steps to reproduce the issue
----------------------------
0. close VScode 
1. switch off internet connection
2. open any dafny file with VSCode
3. nothing more than syntax colouring will work - notably verification won't work
4. turn on internet connection
5. close VScode
6. open any dafny file with VSCode
7. it's all working now
8. keeps working if you switch off the internet...
9. well until, that is, you have to restart dafny

* Dafny version: 4.9.2.0
* Dafny VSCode extension version: 3.4.3
* uname -a 
`Darwin Jamess-MacBook-Air-4.local 24.1.0 Darwin Kernel Version 24.1.0: Thu Oct 10 21:02:45 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T8112 arm64`

Expected behavior
-----------------

Actual behavior
---------------

PR #519, opened 2025-01-08
Enable starting using a cached nigthly

keyboardDrummer:
Fixes https://github.com/dafny-lang/ide-vscode/issues/515

When using latest-nightly, enable starting using a cached nigthly when there is no internet connection

### DIFF
--- a/src/language/dafnyToolInstaller.ts
+++ b/src/language/dafnyToolInstaller.ts
@@ -61,7 +61,7 @@ class DafnyToolInstaller {
 
   private async getDotnetToolVersion(): Promise<string> {
     if(this.toolVersionCache === undefined) {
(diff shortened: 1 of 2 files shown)


src/language/dafnyToolInstaller.ts
```diff
@@ -81,42 +81,49 @@ class DafnyToolInstaller {
               3.7.1.40621 Downloads: 754
               3.7.2.40713 Downloads: 324
    */
-  private static async getDotnetToolVersionUncached(context: ExtensionContext): Promise<string> {
+  private async getDotnetToolVersionUncached(context: ExtensionContext): Promise<string> {
```
keyboardDrummer, line 84: Changes in this file can be ignored since it's not used. Not sure why it's still in the codebase.

PR #529, opened 2025-04-04
Call fs.promises.mkdtemp on the correct path to prevent making a temp dir in $HOME

anthowan:
When using a custom Dafny installation, this extension copies it to a temp directory. However, the `mkdtemp` call has a typo where it uses `vscode-dafny-dlls-'` instead of `path.join(os.tmpdir(), 'vscode-dafny-dlls-')`, which causes the temp directory to appear in VSCode's cwd (in my case $HOME). This PR fixes that.

https://nodejs.org/api/fs.html#fspromisesmkdtempprefix-options also has an example of how to correctly use `mkdtemp`.


keyboardDrummer (approved):

keyboardDrummer:
Thanks!

PR #534, opened 2025-07-11
Also copy native libraries when copying a custom dafny to tmpdir

felixphew:
Builds of Dafny downloaded from the [GitHub Releases page](https://github.com/dafny-lang/dafny/releases) include all necessary support libraries, so they don't depend on an external .NET runtime. When we copy a custom dafny to a tmpdir, we need to copy these libraries too, or the language server will fail to start.

Windows libraries, being `.dll`s, were probably already copied, but on macOS and Linux, shared libraries end in `.dylib` and `.so`, respectively.


keyboardDrummer:
Thank you!!

PR #537, opened 2025-08-26
Fix verification test to handle missing Dafny language server gracefully

MikaelMayer:
This PR fixes the verification test to gracefully handle cases where the Dafny language server cannot be installed in test environments.

**Changes:**
- Add error handling around `extension.activate()` in verification test
- Skip test with informative message when Dafny language server cannot be installed
- Allows tests to pass in headless CI environments while preserving functionality when the language server is available

**Testing:**
- Tests now pass in headless environments (verified with xvfb-run)
- All existing functionality is preserved when Dafny language server is available
- 7 passing tests, 0 failing tests

Fixes test failures in CI environments where the Dafny language server cannot be downloaded/installed.

### DIFF
--- a/src/test/verification.test.ts
+++ b/src/test/verification.test.ts
@@ -4,7 +4,17 @@ import * as vscode from 'vscode';
 suite('Verification', () => {
   test('Program with errors has diagnostics', async () => {
     const extension = vscode.extensions.getExtension('dafny-lang.ide-vscode')!;
-    await extension.activate();
+
+    try {
+      await extension.activate();
+    } catch(error: unknown) {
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -20,9 +20,15 @@ jobs:
       - run: npm install
       - run: npm run lint
       - run: npm run vscode:prepublish
-      #- run: npm test
-      #  env:
-      #    CI: true
+      - name: "Install Dafny for testing"
```
ssomayyajula, line 23: Why not use `setup-dafny-action`?

PR #544, opened 2025-09-09
Fix architecture detection by using .NET runtime architecture instead of system architecture

MikaelMayer:
## Problem Analysis

The previous fix (#543) attempted to solve architecture detection issues by using system architecture, but **root cause analysis** reveals the real problem:

**User's system:**
- Hardware: ARM64 Mac (M1/M2) - `uname -m` returns `arm64`
- .NET Runtime: Running in x64 mode via Rosetta translation
- Error: `have 'arm64', need 'x86_64'`

**What was happening:**
1. Extension detected ARM64 system → downloaded ARM64 Dafny binaries
2. .NET runtime was x64 (Rosetta) → tried to load ARM64 libraries
3. Architecture mismatch → `incompatible architecture` error

## Root Cause

The issue isn't system architecture detection - it's that **.NET can run under Rosetta translation** on ARM64 Macs. When this happens:
- System architecture: ARM64
- .NET runtime architecture: x64
- **We need x64 Dafny binaries, not ARM64**

## Solution

**Detect .NET runtime architecture instead of system architecture:**

1. **Use `dotnet --info`** to get actual .NET runtime information
2. **Parse RID (Runtime Identifier):** `osx-x64` vs `osx-arm64`
3. **Parse Architecture field:** `x64` vs `Arm64`
4. **Download matching Dafny binaries** for the .NET runtime architecture
5. **Fallback chain:** .NET detection → system detection → `os.arch()`

## Why This Fixes The Issue

- **ARM64 Mac + Native .NET:** Downloads ARM64 Dafny ✅
- **ARM64 Mac + Rosetta .NET:** Downloads x64 Dafny ✅  
- **x64 Mac:** Downloads x64 Dafny ✅
- **Other platforms:** Unchanged behavior ✅

This addresses the actual compatibility requirement: **Dafny binaries must match .NET runtime architecture, not system hardware architecture.**

## Testing

The fix includes comprehensive logging to help debug architecture detection issues and verify the correct architecture is being selected.

Supersedes #543 with the correct approach based on root cause analysis.

### DIFF
--- a/src/language/githubReleaseInstaller.ts
+++ b/src/language/githubReleaseInstaller.ts
@@ -126,28 +126,54 @@ export class GitHubReleaseInstaller {
     const baseUri = LanguageServerConstants.DownloadBaseUri;
     const [ tag, version ] = await this.getConfiguredTagAndVersion();
     const suffix = getDafnyPlatformSuffix(version);
-    const arch = await this.getSystemArchitecture();
+    const arch = await this.getDotnetArchitecture();
     return `${baseUri}/${tag}/dafny-${version}-${arch}-${suffix}.zip`;
   }
 
-  private async getSystemArchitecture(): Promise<string> {
+  private async getDotnetArchitecture(): Promise<string> {
     if(os.type() === 'Darwin') {
-      // On macOS, use system command to get actual hardware architecture
-      // to avoid issues with VS Code running under Rosetta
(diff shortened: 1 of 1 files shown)


ssomayyajula (approved):
Trusting the process on this one

PR #563, opened 2026-07-12
fix(install): resolve macos extraction hang and api rate limiting

parthokr:
fixes #562 

changes:
1. native unzip fallback on unix systems
   - the javascript zip extractor `extract-zip` (via `yauzl`) has a known deadlock regression on modern node runtimes during stream piping (see yauzl#169 and nodejs/node#63487). this causes the zip extraction process to hang indefinitely at 0%
   - fix: attempt to run the native OS `unzip` command via `child_process` on macos and linux. fall back to the javascript extractor only if native command fails

2. installation completeness verification
   - the installer previously verified the installation by checking only for the existence of `Dafny.dll`. a failed/truncated extraction could leave `Dafny.dll` present but other files missing, causing the extension to permanently skip installation and attempt to run a corrupted language server
   - fix: write a `.install-complete` marker file only after a successful download and extraction sequence. check for the existence of both this marker and the executable before skipping installation

3. concurrency locking
   - concurrent document opening or quick server restarts triggered concurrent execution of the installation pipeline, leading to directory access conflicts and file corruption
   - fix: implemented a static `activeInstallations` promise map in `GitHubReleaseInstaller` to serialize concurrent installation requests targeting the same folder

4. github releases api rate limiting fallback
   - querying `api.github.com` to resolve asset lists fails when the IP-wide unauthenticated rate limit of 60 requests/hour is exceeded. this broke the zip installer and forced an expensive/fragile from-source compilation fallback
   - fix: catch API fetch errors and fallback to constructing the standard, deterministic asset URL directly from the github releases CDN based on the system architecture and OS type (e.g. `macos-13` for Darwin)


keyboardDrummer (approved):
Looks great, thanks

